PR #7, opened 2018-06-21
Implement in-memory cache and offline caching.

HofmannZ:
* Implemented in-memory cache.
* Write memory to file when in background.
* Added provider widget to save and restore the in-memory cache.
* Restructure the project.

Special thanks to @eusdima for his work on the cache. 🚀


kunaldodiya:
how to user client.cache.read() or client.cache.write() ?? what i am suppose to send as key ???

PR #8, opened 2018-06-22
Explicitly state Map types in RunMutation

camuthig:
The new `RunMutation` type declaration inferencing will default to `<dynamic, dynamic>` without explicitly declaring the expectation. Because of this, the `Mutation` widgets fail to build with: 

`type '(Map<String, dynamic>) => void' is not a subtype of type '(Map<dynamic, dynamic>) => void'`

The stacktrace shows the error at 

`MutationState.build (package:graphql_flutter/src/mutation.dart:63:7)`

I believe the input should always be a `String` indexed map for the variables, so changing the type on the typedef from `Map` to `Map<String, dynamic>` should be fine.

### DIFF
--- a/lib/src/mutation.dart
+++ b/lib/src/mutation.dart
@@ -2,7 +2,7 @@ import 'package:flutter/widgets.dart';
 
 import './client.dart';
 
-typedef void RunMutation(Map variables);
+typedef void RunMutation(Map<String, dynamic> variables);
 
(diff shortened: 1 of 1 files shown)


eusdima:
Thanks for the bug report.
I have already fixed it in the last update that comes with offline caching bashed on hashmap that I am going  to push in a few hours.

PR #13, opened 2018-06-24
Add onCompleted to the Mutation widget

camuthig:
This pull request introduces the `onCompleted` concept from the [Apollo client](https://www.apollographql.com/docs/react/essentials/mutations.html#props). The major use I have found for this so far has been for using the navigator to go to new pages or display alert dialogs after completing the mutation, as demonstrated in the example project.

### DIFF
--- a/example/app/lib/main.dart
+++ b/example/app/lib/main.dart
@@ -74,6 +74,24 @@ class _MyHomePageState extends State<MyHomePage> {
 
               return new Mutation(
                 mutations.addStar,
+                onCompleted: (Map<String, dynamic> data) {
+                  showDialog(
+                    context: context,
+                    builder: (BuildContext context) {
+                      return AlertDialog(
+                        title: Text('Thanks for your star!'),
+                        actions: <Widget>[
+                          SimpleDialogOption(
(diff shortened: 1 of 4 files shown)


example/app/pubspec.yaml
```diff
@@ -8,7 +8,8 @@ dependencies:
   # The following adds the Cupertino Icons font to your application.
   # Use with the CupertinoIcons class for iOS style icons.
   cupertino_icons: ^0.1.2
-  graphql_flutter: ^0.4.1
+  graphql_flutter:
+    path: ../..
```
HofmannZ: Was already looking for a way to do relative dependencies in Flutter, thanks!

example/app/lib/main.dart
```diff
@@ -74,6 +74,24 @@ class _MyHomePageState extends State<MyHomePage> {
 
               return new Mutation(
                 mutations.addStar,
+                onCompleted: (Map<String, dynamic> data) {
+                  showDialog(
```
HofmannZ: Can you move the `onCompleted` below the `builder argument, so it matches the order in the class constructer.

PR #19, opened 2018-06-25
Ignore remaining idea files and fix example pubspec path

camuthig:
I know these two concepts are not generally related, but they were both very simple. Willing to split them up if you would prefer.


HofmannZ:
I held back on those remaining files, because they don't reference any local files. Is there any specific reason why they should not be kept in version control?

> I use VS Code myself so I'm a little bit cautious removing files for other IDE.

camuthig:
I removed them at the suggestion of the IntelliJ documentation: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems .

HofmannZ:
Thanks for the link, makes sense to remove it since it's recommended.

PR #20, opened 2018-06-25
Update iOS project files

camuthig:
I'm not actually sure if the two ios files need to be updated/committed, as I'm not sure if this is part of my system definition or part of initializing the iOS project - it is a little difficult to tell. According to the Flutter team, both files are important, and not part of their gitignore for that reason. The update to the .gitignore is to ensure the Podfile.lock is ignored, as suggested by the Flutter team.


HofmannZ:
LGTM

PR #22, opened 2018-06-27
Authorization Header

apexJCL:
This prevents from sending a null Auth header (for troublesome endpoints), also allows modifying the prefix, if anyone is using other than the usual `Bearer`

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,8 +51,8 @@ void main() async {
     cache: new InMemoryCache(), // currently the only cache type we have implemented.
   );
   client.apiToken = '<YOUR_GITHUB_PERSONAL_ACCESS_TOKEN>';
-  client.apiTokenPrefix = 'JWT'; // Default is 'Bearer'
-
+  // If you need a different auth token format, use getAuthorizationHeader callback function
+  client.getAuthorizationHeader = (String apiToken) => '<YOUR_CUSTOM_AUTHORIZATION_STRING>'
   ...
 }
 
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ void main() async {
     cache: new InMemoryCache(), // currently the only cache type we have implemented.
   );
   client.apiToken = '<YOUR_GITHUB_PERSONAL_ACCESS_TOKEN>';
+  client.apiTokenPrefix = 'JWT'; // Default is 'Bearer'
 
   ...
 }
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -12,7 +12,6 @@ class Client {
     String endPoint = '',
     String apiToken,
     InMemoryCache cache,
-    Map<String, String> Function(Map<String, String>) handleHeaders,
     String Function(String) getAuthorizationHeader,
   }) {
     this.endPoint = endPoint;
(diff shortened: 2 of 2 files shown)


lib/src/client.dart
```diff
@@ -37,15 +39,25 @@ class Client {
...
   String get apiToken => this._apiToken;
 
-  Map<String, String> get headers => {
-        'Authorization': 'Bearer $apiToken',
-        'Content-Type': 'application/json',
-      };
+  String get apiTokenPrefix => this._apiTokenPrefix;
+
+  Map<String, String> get headers {
+    Map<String, String> h = new Map();
```
eusdima: Rename h to a descriptive name such as `headersMap `

Add content-type to the initialisation 
`Map<String, String> headersMap = { 'Content-Type':  'application/json'}`

New line in between variable declaration and if statement to separate code blocks

lib/src/client.dart
```diff
@@ -37,15 +39,25 @@ class Client {
...
+    if (apiToken != null) {
+      h['Authorization'] = '$apiTokenPrefix $apiToken';
```
eusdima: Now you are fixing the issue where the user has a different `apiTokenPrefix`, but what happens when the `Authorization ` has a different format?

Maybe allow the user to define the getter function for Authorization header, therefore removing the need for `apiTokenPrefix ` when the user defines his own function, or uses `'Bearer'` by default.

eusdima: That middleware adds unnecessary overhead to every request.

README.md
```diff
@@ -51,6 +51,7 @@ void main() async {
     cache: new InMemoryCache(), // currently the only cache type we have implemented.
   );
   client.apiToken = '<YOUR_GITHUB_PERSONAL_ACCESS_TOKEN>';
+  client.apiTokenPrefix = 'JWT'; // Default is 'Bearer'
```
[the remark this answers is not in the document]
HofmannZ: Not sure if overwriting the meths is the best way to do it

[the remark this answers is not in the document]
eusdima: You re not overwriting, you are defining a new method if you want custom auth headers

HofmannZ:
@apexJCL, @eusdima, I actually think it would be more flexible to setup something like [Apollo](https://www.apollographql.com/docs/react/recipes/authentication.html) did. What do you guys think about that?

apexJCL:
@HofmannZ that was my __intented__ prev commit, where we declare a middleware to handle it [like they do over here)[https://www.apollographql.com/docs/angular/recipes/authentication.html#Header], maybe my implementation was wrong, but the idea was to do it like that.

HofmannZ:
@apexJCL I think the best way to mimic that is to create a new class for `HttpLink`, and add the `setContext` as a method on that class. What do you think?

Also we've been using React-Apollo as reference for the project, so maybe it's a good idea to not mix Angular with it ;)

apexJCL:
@HofmannZ Then maybe it'd be better to go full on the approach of `links`, so we can define whether the connection is made throught `http` or maybe a `websocket`, allowing for that kind of context handling.

PS: I didn't realize it was the Angular doc, but the concept it's pretty much the same (as far as I can see)

eusdima:
I would leave Zino in charge of this because he has a better understanding of the roadmap following apollo client.

HofmannZ:
@apexJCL Started on the full approach with links, see #36

Issue #16, opened 2018-06-25
Pass Client object as parameter, or let client object define query methods


ened:
Currently, I have code like this:

*lib/main.dart:*
```
void main() {
  client = new Client(
    endPoint: 'https://api.github.com/graphql',
    cache: new InMemoryCache(),
  );
  client.apiToken = "...";

  runApp(new MyApp());
}
```

And somewhere in the app, I will then simply run my query like so:
```
return Query(
      readRepositories,
...
)
```

This is very simple, client & query are loosely coupled.
The `client` variable is actually a global.

I think it would be preferable to not have a global client object, and instead have it available as a (named) parameter to the Query or other objects.

Client could also define `newQuery` or `newMutation` methods for convenience.

HofmannZ:
@ened Thanks for the suggestion. @eusdima I think we discussed this internally.

Both solutions don't seem ideal. Providing the client with each query or mutation is too cumbersome. And a global client variable is not useful if you might want connect to multiple endpoints.

Instead I propose to create a `GraphqlProvider` widget that takes a client as a param, and exposes it to all its children trough the context. This would mimic the behaviour of the [ApolloProvider](https://www.apollographql.com/docs/react/api/react-apollo.html#ApolloProvider)

What do you think?

eusdima:
Creating a provider and passing client as context seems good, but you will not be able to use the client outside of the provider so maybe support both: GraphqlProvider and the old way.

We have to do performance measurement to make sure that the memory and cpu usage are not higher after switching to the `GraphqlProvider`.

So should we keep both or just  `GraphqlProvider`?

HofmannZ:
Well you can always declare the client somewhere in `main.dart` then pass that client to the `GraphqlProvider`.  Now you can still use the client variable directly as needed.

We can also use [ApolloConsumer](https://www.apollographql.com/docs/react/api/react-apollo.html#apollo-consumer) to acces the compete client within the context tree.

HofmannZ:
Lets do that, I’ll start on a PR later today. We have to setup some test to track the performance impact on the queries. ✌🏼

HofmannZ:
👆🏼 Early feedback is much appreciated

HofmannZ:
@ened,
I just finished PR #24, it implements the `GraphqlProvider` & `GraphqlConsumer`. I though you might want to have a look and give some feedback, since you brought the issue to light.

ened:
[...]

PR #24, opened 2018-06-28
`GraphqlProvider` and `GraphqlConsumer` widgets

HofmannZ:
This PR adds the `GraphqlProvider` and `GraphqlConsumer` widgets as discussed in #16.

### Breaking change

- The library now requires your app to be wrapped with the `GraphqlProvider` widget @HofmannZ
- The global `client` variable is no longer available. Instead use the `GraphqlConsumer` widget

#### Fixes / Enhancements

- Added the `GraphqlProvider` widget. The client is now stored in an `InheritedWidget`, and can be accessed anywhere within the app.

```dart
Client client = GraphqlProvider.of(context).value;
```

- Added the `GraphqlConsumer` widget. For ease of use we added a widget that uses the same builder structure as the `Query` and `Mutation` widgets. 

> Under the hood it access the client from the `BuildContext`.

- Added the option to optionally provide the `apiToken` to the `Client` constructor. It is still possible to set the `apiToken` with setter method.

```dart
  return new GraphqlConsumer(
    builder: (Client client) {
      // do something with the client

      return new Container();
    },
  );
```

#### Docs

- Added documentation for the new `GraphqlProvider`
- Added documentation for the new `GraphqlConsumer`
- Changed the setup instructions to include the new widgets
- Changed the example to include the new widgets

### DIFF
--- a/lib/src/widgets/query.dart
+++ b/lib/src/widgets/query.dart
@@ -33,12 +33,14 @@ class QueryState extends State<Query> with WidgetsBindingObserver {
   Map<String, dynamic> data = {};
   String error = '';
   Duration pollInterval;
+  Timer pollTimer;
+  bool initialFetch = false;
 
   @override
   void initState() {
     super.initState();
 
-    if (widget.pollInterval != null) {
+    if (widget.pollInterval is int) {
       pollInterval = new Duration(seconds: widget.pollInterval);
     }
   }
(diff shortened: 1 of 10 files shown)


lib/src/widgets/query.dart
```diff
@@ -33,12 +33,14 @@ class QueryState extends State<Query> with WidgetsBindingObserver {
   Map<String, dynamic> data = {};
   String error = '';
   Duration pollInterval;
+  Timer pollTimer;
+  bool initialFetch = false;
```
HofmannZ: Default to true

lib/src/widgets/query.dart
```diff
@@ -92,9 +94,11 @@ class QueryState extends State<Query> with WidgetsBindingObserver {
   }
 
   Widget build(BuildContext context) {
-    final Client client = GraphqlProvider.of(context).client;
+    if (!initialFetch) {
+      initialFetch = true;
```
HofmannZ:
```dart
if (initialFetch) {
  initialFetch = false;
  ....
}
```

PR #32, opened 2018-07-15
Fix mutations example bug not updating star bool

cal-pratt:
Got caught in a few small places while trying to launch the example.

Added a stack trace in `mutation.dart`. The current implementation only outputs the high-level description making it difficult to track where the exception is coming from.

Exception logic was throwing in `client.dart`. It is possible for `jsonResponse['errors']` to be a `List<dynamic>` instead of a `String`

Update the local data in the Mutation Widget based on the result of the Mutation. This was causing stars to remain uncolored after performing a tap

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -80,6 +80,11 @@ class _MyHomePageState extends State<MyHomePage> {
                   Map data,
                   String error,
                 }) {
+                  if (data.isNotEmpty) {
+                    repository['viewerHasStarred'] =
+                      data['addStar']['starrable']['viewerHasStarred'];
+                  }
+
                   return new ListTile(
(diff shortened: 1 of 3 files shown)


example/lib/main.dart
```diff
@@ -80,6 +80,11 @@ class _MyHomePageState extends State<MyHomePage> {
                   Map data,
                   String error,
                 }) {
+                  if (data.isNotEmpty) {
+                    repository['viewerHasStarred'] =
+                      data['addStar']['starrable']['viewerHasStarred'];
+                  }
```
HofmannZ: Until we properly restructure the result data, this is a good alternative for the optimistic UI updates.

PR #33, opened 2018-07-16
Subscription SocketClient and Widget

cal-pratt:
Hi Everyone, 

Just want to start off by saying thanks! I think this is an awesome project and am excited to see it grow! For my personal project I need to use GraphQL subscriptions, so I figured I'd try my hand at a proof of concept. When you get some time take a look and let me know if it's in line with how you see Subscriptions working in this project. As an FYI I'm new to dart programming and my style is closer to Java development.. all nitpicks are welcome! :)

I implemented the subscriptions by following this guide from [apollographql/subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md). ~~This work does not yet contain any flutter `Widget` classes, it's just the implementation of the websocket client.~~

Because a lot of the parameters for the `http` and `ws` endpoints can vary pretty heavily, I decided to implement a separate `socket_client` instead of merging it with the existing http `client`. I also chose to wrap the data types being sent over the wire as concrete dart classes as opposed to just using `dynamic`. Not sure if this is the best solution as translating from dart object to json and vice versa doesn't seem to be totally trivial in dart... that said you do get better compile time safety from it. I chose to also implement this solution using the basic `Stream` api, but I feel we might get a lot of benefit from introducing `rxdart` into the project. 

I ran the following command-line program against my graphql server running on graphql-dotnet. I don't believe the github api has any available subscription api's to use in the example program. It might be interesting to host a dedicated server somewhere for handling development requests.

```dart
main() {
  _main();
}

_main() async {
  socketClient = await SocketClient.connect('ws://192.168.1.102:55855/graphql');

  var operationName = "SessionAudioEvent";
  var query = """subscription $operationName(\$sessionId: String!) {
    sessionAudioEvent(sessionId: \$sessionId) {
      audio {
        volume
      }
    }
  }""";
  var variables = {'sessionId': '5aae762d-cb4d-e909-7424-32251326e18b'};

  socketClient.subscribe(SubscriptionRequest(operationName, query, variables))
      .listen(print);
}
```

Prints the following:

```text
{type: connection_ack}
{type: data, data: {sessionAudioEvent: {audio: {volume: 0.8765849471092224}}}, errors: {}}
{type: data, data: {sessionAudioEvent: {audio: {volume: 0.6523423194885254}}}, errors: {}}
{type: data, data: {sessionAudioEvent: {audio: {volume: 0.38732823729515076}}}, errors: {}}
```

**Update:**

Following the implementation of the `Mutation` widget, I was able to implement a `Subscription` widget. One of the tougher parts of creating this widget is making sure that the subscription object is only constructed once. To do this I override the `initState` and `dispose` methods on the widget `State` class. With the following modification to the example project `main.dart` file (and the constructs created in the above example) I was able to implement a real-time `Widget`:

```dart 

  @override
    return new Scaffold(
      body: new Center(
        child: new Subscription(SubscriptionRequest(operationName, query, variables),
          builder: ({bool loading, SubscriptionData payload, SubscriptionError error}) {
            if (payload?.data != null) {
              var volume = payload.data['sessionAudioEvent']['audio']['volume'] ?? 0.0;
              return new Slider(value: volume, onChanged: (v) {});
            } else {
              return new Slider(value: 0.0, onChanged: (v) {});
            }
          }
        ),
      )
    );
  }
}
```

### DIFF
--- a/lib/src/socket_client.dart
+++ b/lib/src/socket_client.dart
@@ -1,0 +1,64 @@
+import 'dart:async';
+import 'dart:io';
+
+import 'package:uuid/uuid.dart';
+
+import '../graphql_flutter.dart';
+
+SocketClient socketClient;
+
+class SocketClient {
+  static Future<SocketClient> connect(
+    final String endPoint, {
+    final List<String> protocols = const [
+      'graphql-ws',
+    ],
+    final Map<String, String> headers = const {
+      'content-type': 'application/json',
+    },
+  }) async {
+    return SocketClient(GraphQLSocket(await WebSocket.connect(
+      endPoint,
+      protocols: protocols,
+      headers: headers,
+    )));
+  }
+
+  final Uuid _uuid = Uuid();
+  final GraphQLSocket _socket;
+
+  SocketClient(this._socket) {
+    _socket.connectionAck.listen(print);
+    _socket.connectionError.listen(print);
+    _socket.unknownData.listen(print);
+    _socket.write(InitOperation());
+  }
+
+  Stream<SubscriptionData> subscribe(final SubscriptionRequest payload) {
+    final String id = _uuid.v4();
+
+    final StreamController<SubscriptionData> response =
+        StreamController<SubscriptionData>();
+
+    final Stream<SubscriptionComplete> complete = _socket.subscriptionComplete
+        .where((message) => message.id == id)
+        .take(1);
+
+    final Stream<SubscriptionData> data = _socket.subscriptionData
+        .where((message) => message.id == id)
+        .takeWhile((_) => !response.isClosed);
+
+    final Stream<SubscriptionError> error = _socket.subscriptionError
+        .where((message) => message.id == id)
+        .takeWhile((_) => !response.isClosed);
+
+    complete.listen((_) => response.close());
+    data.listen((message) => response.add(message));
+    error.listen((message) => response.addError(message));
+
+    response.onListen = () => _socket.write(StartOperation(id, payload));
+    response.onCancel = () => _socket.write(StopOperation(id));
+
+    return response.stream;
+  }
+}
--- a/lib/src/websocket/messages.dart
+++ b/lib/src/websocket/messages.dart
@@ -1,0 +1,195 @@
+/// These messages represent the structures used for Client-server communication
+/// in a GraphQL web-socket subscription. Each message is represented in a JSON
+/// format where the data type is denoted by the `type` field.
+
+/// A list of constants used for identifying message types
+class MessageTypes {
+  MessageTypes._();
+
+  // client connections
+  static const GQL_CONNECTION_INIT = 'connection_init';
+  static const GQL_CONNECTION_TERMINATE = 'connection_terminate';
+
+  // server connections
+  static const GQL_CONNECTION_ACK = 'connection_ack';
+  static const GQL_CONNECTION_ERROR = 'connection_error';
+
+  // client operations
+  static const GQL_START = 'start';
+  static const GQL_STOP = 'stop';
+
+  // server operations
+  static const GQL_DATA = 'data';
+  static const GQL_ERROR = 'error';
+  static const GQL_COMPLETE = 'complete';
+
+  // default tag for use in identifying issues
+  static const GQL_UNKNOWN = 'unknown';
+}
+
+abstract class JsonSerializable {
+  dynamic toJson();
+
+  @override
+  String toString() => toJson().toString();
+}
+
+/// Base type for representing a server-client subscription message.
+abstract class GraphQLSocketMessage extends JsonSerializable {
+  final String type;
+
+  GraphQLSocketMessage(this.type);
+}
+
+/// After establishing a connection with the server, the client will
+/// send this message to tell the server that it is ready to begin sending
+/// new subscription queries.
+class InitOperation extends GraphQLSocketMessage {
+  InitOperation() : super(MessageTypes.GQL_CONNECTION_INIT);
+
+  @override
+  dynamic toJson() => {
+    'type': type,
+  };
+}
+
+/// Represent the payload used during a Start query operation.
+/// The operationName should match one of the top level query definitions
+/// defined in the query provided. Additional variables can be provided
+/// and sent to the server for processing.
+class SubscriptionRequest extends JsonSerializable {
+  final String operationName;
+  final String query;
+  final dynamic variables;
+
+  SubscriptionRequest(this.operationName, this.query, this.variables);
+
+  @override
+  dynamic toJson() => {
+    'operationName': operationName,
+    'query': query,
+    'variables': variables,
+  };
+}
+
+/// A message to tell the server to create a subscription. The contents of the
+/// query will be defined by the payload request. The id provided will be used
+/// to tag messages such that they can be identified for this subscription
+/// instance. id values should be unique and not be re-used during the lifetime
+/// of the server.
+class StartOperation extends GraphQLSocketMessage {
+  final String id;
+  final SubscriptionRequest payload;
+
+  StartOperation(this.id, this.payload) : super(MessageTypes.GQL_START);
+
+  @override
+  dynamic toJson() => {
+    'type': type,
+    'id': id,
+    'payload': payload,
+  };
+}
+
+/// Tell the server to stop sending subscription data for a particular
+/// subscription instance. See StartOperation
+class StopOperation extends GraphQLSocketMessage {
+  final String id;
+
+  StopOperation(this.id) : super(MessageTypes.GQL_STOP);
+
+  @override
+  dynamic toJson() => {
+    'type': type,
+    'id': id,
+  };
+}
+
+/// The server will send this acknowledgment message after receiving the init
+/// command from the client if the init was successful.
+class ConnectionAck extends GraphQLSocketMessage {
+  ConnectionAck() : super(MessageTypes.GQL_CONNECTION_ACK);
+
+  @override
+  dynamic toJson() => {
+    'type': type,
+  };
+}
+
+/// The server will send this error message after receiving the init command
+/// from the client if the init was not successful.
+class ConnectionError extends GraphQLSocketMessage {
+  final dynamic payload;
+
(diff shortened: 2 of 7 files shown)


HofmannZ:
Thanks for the really well structured PR.

We are planning to split the links from the client into their own classes and ideally into their own libraries. Right now we only had the HTTP link so the WS link would be a good incentive to start splitting some the code base.

The `Subscription` widget we should be fairly similar to the `Mutation` with the exception of a `shouldResubscribe` option.

I'm currently traveling so I have limited time to write code the coming two weeks, but will try to squeeze some time in, because this PR is an exciting addition.

cal-pratt:
Hey so I looked into the `Mutation` `Widget` a bit more. Once I got my head around the `typedef` syntax everything became much more clear and I was able to make a working `Widget` POC!! I'll update the main PR description.

lib/src/socket_client.dart
```diff
@@ -0,0 +1,47 @@
...
+
+SocketClient socketClient;
+
+class SocketClient {
+  static Future<SocketClient> connect(final String endPoint, {
+    final List<String> protocols = const ['graphql-ws'],
+    final Map<String, String> headers = const {'Content-Type': 'application/json'}
+  }) async {
+    return SocketClient(GraphQLSocket(await WebSocket.connect(
+        endPoint, protocols: protocols, headers: headers)));
```
eusdima: comma after `headers: headers`

lib/src/socket_client.dart
```diff
@@ -0,0 +1,47 @@
...
+
+  SocketClient(this._socket) {
+    _socket.connectionAck.listen(print);
+    _socket.connectionError.listen(print);
+    _socket.unknownData.listen(print);
+    _socket.write(InitOperation());
+  }
+
+  Stream<SubscriptionData> subscribe(final SubscriptionRequest payload) {
+    var id = _uuid.v4();
```
eusdima: Change var to a more specific type

lib/src/socket_client.dart
```diff
@@ -0,0 +1,47 @@
...
+
+    var response = new StreamController<SubscriptionData>();
+
+    var complete = _socket.subscriptionComplete.where((m) => m.id == id).take(1);
```
eusdima: rename `m` to a more specific name

lib/src/websocket/messages.dart
```diff
@@ -0,0 +1,141 @@
...
+}
+
+abstract class JsonSerializable {
+  dynamic toJson();
+  @override toString() => toJson().toString();
+}
+
+/// Base type for representing a server-client subscription message.
+abstract class GraphQLSocketMessage extends JsonSerializable {
+  final String type;
```
eusdima: Newline after variable declaration

lib/src/websocket/messages.dart
```diff
@@ -0,0 +1,141 @@
...
+  GraphQLSocketMessage(this.type);
+}
+
+/// After establishing a connection with the server, the client will
+/// send this message to tell the server that it is ready to begin sending
+/// new subscription queries.
+class InitOperation extends GraphQLSocketMessage {
+  InitOperation() : super(MessageTypes.GQL_CONNECTION_INIT);
+
+  @override toJson() => {'type': type};
```
eusdima: comma after `'type': type`

lib/src/widgets/subscription.dart
```diff
@@ -0,0 +1,77 @@
...
+    stream.takeWhile((m) => this._alive)
+        .listen(_onData, onError: _onError, onDone: _onDone);
+
+    if (widget.initial != null) {
+      _onData(widget.initial);
+    }
+  }
+
+  @override dispose() {
+    super.dispose();
```
eusdima: put `super.dispose` at the end

lib/src/socket_client.dart
cal-pratt: Are you sure we want to enforce trailing comma for one-line declarations?
I understand the motivation for all commas on multi-line assignments, but the syntax definitely seems odd for when it is adjacent to a brace. Maybe we'd have to follow this up with an *always chop* or *closing brace on new line*

eg) 
```dart
return SocketClient(GraphQLSocket(await WebSocket.connect(
  endPoint, protocols: protocols, headers: headers,
)));
```
or
```dart
return SocketClient(GraphQLSocket(await WebSocket.connect(
  endPoint, 
  protocols: protocols, 
  headers: headers,
)));
```

lib/src/socket_client.dart
cal-pratt: In this case I think I agree that `var` is better. The method `Uuid::v4` returns a dynamic type, and denotes it may not always be a string depending on the arguments provided. Perhaps it would be better to explicitly call `toString()` on it as well for safety... 

For other cases I'd like to debate the comment you made:  

> "give types to variables not just 'var' (you can just hover over the method or variable and it will you give the return type":

In the official [dart style guide](https://www.dartlang.org/guides/language/effective-dart/design#avoid-type-annotating-initialized-local-variables) they specifically say to prefer `var` for local scoped variables. Type inference works pretty well in the local scopes, and leaves the denser logic less cluttered. It seems like a waste of the language feature to not embrace it.

> Local variables, especially in modern code where functions tend to be small, have very little scope. Omitting the type focuses the reader’s attention on the more important name of the variable and its initialized value.

To that point I'm open to any styling rules as long as we can stay consistent throughout the project. I'll start looking to see if there are any good static analysis/ checkstyle tools we can apply to the build such that we can begin codifying these style issues!

cal-pratt: Ah, on my second search I see that flutter has their own style guidelines that contradicts the dart-lang crew: 

> Always avoid "var". Use "dynamic" if you are being explicit that the type is unknown. Use "Object" if you are being explicit that you want an object that implements == and hashCode.

https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-using-var

Seems rather restrictive to me, but I guess staying consistent with the framework is more useful here?

¯\\_(ツ)_/¯

lib/src/socket_client.dart
cal-pratt: Ah, flutter actually backs me up this time 😃 

> Use a trailing comma for arguments, parameters, and list items, but only if they each have their own line.

https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#use-a-trailing-comma-for-arguments-parameters-and-list-items-but-only-if-they-each-have-their-own-line

lib/src/widgets/subscription.dart
cal-pratt: Nice catch!

lib/src/socket_client.dart
HofmannZ: @cal-pratt We use format on save, which automatically puts the code on multiple lines if there is a tailing comma. That's why we can always enforce the tailing comma, and still comply with the style guide ;)

Besides that, is's the preferred way to write widgets: [https://flutter.io/formatting/#using-trailing-commas](https://flutter.io/formatting/#using-trailing-commas).

lib/src/socket_client.dart
HofmannZ: Well they are quite contradicting, I think it's best to pick one for this repo and stick with it. Personally the [Flutter style guide](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo) seems like a more appropriate solution for this project, because the project is a Flutter library. But feel free to change my mind.

cal-pratt: Yeah I agree with using the flutter spec for now! In the future if the project splits into two packages then perhaps we should follow dart-lang for the base package and flutter for the Widgets package.

lib/src/socket_client.dart
cal-pratt: Cool. I'm going to take the time and go through the flutter spec to clean up this code. 
Should have a new revision up sometime this weekend

lib/src/socket_client.dart
HofmannZ: Yeah good idea

HofmannZ:
I just finished the PR for the GraphqlProvider #24. There I introduced a small breaking change; the client is no longer available as a global variable. We will merge the change somewhere today or tomorrow.

I thought I would give you a quick heads up. The changes required are really small anyway:
When you need to access the client just use this:
```dart
Client client = GraphqlProvider.of(context).value;
```
and be sure to import `GraphqlProvider` from `src/widgets/graphql_provider.dart`

✌🏻

cal-pratt:
@HofmannZ That sounds good, I always try to rebase often. Do you want me to update the provider with the socket client in the PR? Or would you rather we keep it global for now and make that change in another pull?

HofmannZ:
@cal-pratt Ideally we would setup the links, where we have a socket link and a http link. Then we can merge the two clients into one which allows us to call `links.http.post` for queries & mutations and `links.socket.connect` for subscriptions.

Because that will be quite a bit of work and there is a load of animo from the community to get subscriptions working, we might want to move that into it's own PR. So you can keep the `socketClient` global for now.

On the other side doing it right away will avoid a load of refactoring, and since I have two free days I might be able to squeeze in some the time to finish it before the weekend.

What do you think?

cal-pratt:
@HofmannZ Whatever works with your schedule. I'm hoping to have this shipped soon, permitting the review is fine after my next update this weekend (hopefully sometime tomorrow or Saturday). So there shouldn't be too much down time. That said, if you go ahead and make the `links` object then I can just submit another CR and add the field to that.

HofmannZ:
@cal-pratt Prefect! Currently working on links. It turns out to be a bit more work then expected, but should be done before the weekend.

cal-pratt:
Okay I believe I've taken care of most of the issues.

 - eliminated `var` in favor of `final <Type>`
 - eliminated `new` calls
 - changed to use chopped down array/dictionary syntax with trailing comma
 - added example usage in READ.ME

cal-pratt:
Also I moved the boxing and unboxing of the `SubscriptionRequest` / `SubscriptionData` out of the `Widget` ctor and `Widget` builder objects. This detail is now an internal implementation detail to the `Widget`.

PR #36, opened 2018-07-21
WIP: Links and Query batching

HofmannZ:
Refactoring the client to accept links, also adding a query manager for poling.

Still work in progress...

### DIFF
--- a/lib/src/links/http/fallback_http_config.dart
+++ b/lib/src/links/http/fallback_http_config.dart
@@ -1,0 +1,19 @@
+Map<String, bool> defaultHttpOptions = {
+  'includeQuery': true,
+  'includeExtensions': false,
+};
+
+Map<String, String> defaultHeaders = {
+  'accept': '*/*',
+  'content-type': 'application/json',
+};
+
+Map<String, String> defaultOptions = {
+  'method': 'POST',
+};
+
+Map<String, Map<String, dynamic>> fallbackHttpConfig = {
+  'http': defaultHttpOptions,
+  'headers': defaultHeaders,
+  'options': defaultOptions,
+};
--- a/lib/src/links/http/link_http.dart
+++ b/lib/src/links/http/link_http.dart
@@ -1,0 +1,171 @@
+import 'dart:async';
+import 'dart:convert';
+
+import 'package:http/http.dart' as http;
+
+import 'package:graphql_flutter/src/links/link.dart';
+import 'package:graphql_flutter/src/links/operation.dart';
+import 'package:graphql_flutter/src/links/http/fallback_http_config.dart';
+
+Map<String, dynamic> _selectHttpOptionsAndBody(
+  Operation operation,
+  Map<String, dynamic> fallbackConfig, [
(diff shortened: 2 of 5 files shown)


lib/src/links/operation.dart
```diff
@@ -0,0 +1,62 @@
+import 'dart:convert';
+
+class Operation {
+  Operation({
```
cal-pratt: should this ctor be private?

lib/src/links/operation.dart
```diff
@@ -0,0 +1,62 @@
...
+  String toKey() {
+    /// XXX we're assuming here that variables will be serialized in the same order.
+    /// that might not always be true
+    String encodedVariables = json.encode(variables);
+
+    return '$query|$encodedVariables|$operationName';
+  }
+}
+
+createOperation(
```
cal-pratt: explicitly state return type or perhaps declare as factory.

lib/src/links/http/fallback_http_config.dart
```diff
@@ -0,0 +1,19 @@
+Map<String, bool> defaultHttpOptions = {
```
cal-pratt: Coding around dynamic seems error prone to me. What do you think about using mutable objects with a basic builder pattern?:

```dart
class HttpOptions {  
  bool includeQuery = true;
  bool includeExtensions = false;
}
// user code:
HttpOptions options = HttpOptions()
  ..includeQuery = true
  ..includeExtensions = true;
```

For objects over the wire, Json serialization can be achieved like in the subscriptions PoC.

lib/src/links/http/link_http.dart
```diff
@@ -0,0 +1,171 @@
...
+      }
+    }
+
+    controller = new StreamController(onListen: onListen);
+
+    return controller.stream;
+  });
+}
+
+class HttpLink extends Link {
```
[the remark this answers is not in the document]
eusdima: @HofmannZ and me usually put private ones first

lib/src/links/http/fallback_http_config.dart
eusdima: @cal-pratt Nice suggestion

lib/src/links/link.dart
```diff
@@ -0,0 +1,47 @@
+import 'dart:async';
+
+import 'package:graphql_flutter/src/links/operation.dart';
+
+typedef Stream<Map<String, dynamic>> NextLink(
```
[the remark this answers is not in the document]
HofmannZ: Nice one, totally missed out on the new typedef syntax. I will refactor refactor the app to get rid of it.

lib/src/links/operation.dart
HofmannZ: Factory might be the way to go

lib/src/links/operation.dart
HofmannZ: If we switch to a factory not

HofmannZ:
@cal-pratt managed to get a couple hours in to work on the new link system. I think including subscriptions won't be to much of a hassle what do you think?

PR #38, opened 2018-07-22
Upgrade Gradle version and build tools

cal-pratt:
The existing version of Gradle does not support Java 10 and will lead to compile errors for JDK 10 users.
Bumping up the version such that java 10 and 8 can be supported.


HofmannZ:
LGTM

PR #46, opened 2018-08-04
GraphqlConsumer missing from library

AleksandarFaraj:
Am I the only one using this library?


HofmannZ:
Thanks for reporting the bug.

After writing the GraphqlConsumer we found a more versatile way of accessing the client trough the context using an InheritedWidget. Therefore the GraphqlConsumer has become obsolete. On the other hands, since it's documented we might as well export it for developer to use.

AleksandarFaraj:
@HofmannZ interesting, got any examples of how to access that?

PR #62, opened 2018-08-20
add error extensions support

### LINKED ISSUE #61: Support custom error response extensions
Our GraphQL API returns customized data in error extensions. This plugin catches errors and parses them into a`GQLException` which includes a `GQLError`. This is convenient but the `GQLError` does not include the extensions object. Apollo documentation here: https://www.apollographql.com/docs/apollo-server/v2/features/errors.html. I need to be able to extract the error information as-is from the library.

Add a simple `Map<String, dynamic> extensions` object to `GQLError` so that the host application can read the data when needed.

I don't know of any alternatives. Extensions are free-form so we can't build in a model witch which to parse them.

**Additional context**
none.

dustin-graham:
Fixes #61 . See Apollo documentation on custom error responses: https://www.apollographql.com/docs/apollo-server/v2/features/errors.html

also, export exceptions file. host applications can't effectively catch errors if they can't see the
types.

### DIFF
--- a/lib/src/exceptions.dart
+++ b/lib/src/exceptions.dart
@@ -25,7 +25,8 @@ class GQLError {
 
   /// The path of the field in error.
   final List<dynamic> path;
-  
+
+  /// Custom error data returned by your GraphQL API server
   final Map<String, dynamic> extensions;
 
   /// Constructs a [GQLError] from a JSON map.
--- a/lib/src/exceptions.dart
+++ b/lib/src/exceptions.dart
@@ -25,13 +25,16 @@ class GQLError {
 
   /// The path of the field in error.
   final List<dynamic> path;
+  
+  final Map<String, dynamic> extensions;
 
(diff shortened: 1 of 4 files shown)


dustin-graham:
@HofmannZ I'm using this PR in my local project and it's doing exactly what I needed it to do. please let me know if you have any questions or concerns.

lib/src/exceptions.dart
```diff
@@ -25,13 +25,16 @@ class GQLError {
 
   /// The path of the field in error.
   final List<dynamic> path;
+  
+  final Map<String, dynamic> extensions;
```
HofmannZ: Plase a comment explaining the variable on the  line directly above the declaration. Most editors can process those comments and automatically generate documentation from it.

dustin-graham:
@HofmannZ posted updates. does it look good?

HofmannZ:
LGTM

PR #63, opened 2018-08-21
Breaking Change: Return full GraphQL errors in the responses

### LINKED ISSUE #51: Data from partially successful queries is not returned
**Describe the bug**
If a named query contains an error, data from successfully executed subqueries is simply discarded.  

I have a similar query to this:
```graphql
query getStats {
  statusA: getSystemStatusA {
    uptime
    cpuLoad
  }
  statusB: getSystemStatusB {
    uptime
    cpuLoad
  }
  statusC: getSystemStatusC {
    uptime
    cpuLoad
  }
}
```
which returns this result:
```json
{
  "errors": [
    {
      "message": "error message",
      "locations": [
        {
          "line": 10,
          "column": 3
        }
      ],
      "path": [
        "statusB"
      ]
    },
  ],
  "data": {
    "statusA": {
      "uptime": 123,
      "cpuLoad": 6.25,
    },
    "statusB": null,
    "statusC": {
      "uptime": 456,
      "cpuLoad": 6.25,
    }
  }
}
```
**Expected behavior**
I'd expect to receive the errors and the data without an exception. I'd only expect an exception if the http status code is not 200.

**Additional context**
Problem is a bit of code in [client.dart[78]](https://github.com/zino-app/graphql-flutter/blob/b481f36d4ece22bd899a069b73d8df3c78e22e06/lib/src/client.dart#L78)

Is there any special reason to drop all the data? I'd like to show an appropriate error message in my app ui for the failed queries. 

My current simplistic solution is to comment out the if block and just return the whole JSON object, including all errors.
```Dart
    final Map<String, dynamic> jsonResponse = json.decode(response.body);
    return jsonResponse;
    
    /*     
    if (jsonResponse['errors'] != null && jsonResponse['errors'].length > 0) {
      throw GQLException(
        'Error returned by the server in the query',
        jsonResponse['errors'],
      );
    } 

    return jsonResponse['data'];
    */
```
Thanks!

fusion44:
Hi all!

This fixes #51. This is a breaking change since the full response is now returned instead only the data part. Before this change, if there was a GraphQL Error, it'd always throw a GraphQLError and discard all correct data in the response.


HofmannZ:
The query and mutation widgets don't split the data from the errors yet, therefore the error field is always null. Also make sure all commits are GPG signed.

HofmannZ:
@mmadjer Hey man, you dropped in the error handling. We'll be reviewing this PR ourselves as well, but good to see your opinion on the changes.

mmadjer:
I think that sounds good, @HofmannZ.

fusion44:
Solving this with streams sounds good to me. I'm on vacation this week, so I'll not be able to work on this until I return.

Issue #60, opened 2018-08-20
Handle errors even when no location is defined 


adelcasse:
**Describe the bug**

When we receive errors from a GraphQL servers, the library builds a GQLError : 

  GQLError.fromJSON(Map data)
      : message = data['message'],
        locations = new List.from(
            (data['locations']).map((d) => new Location.fromJSON(d))),
        path = data['path'];

As it is defined, this constructor will fail if the jsonResponse doesn't include a location.

But as far as I understand the GraphQL specification ( http://facebook.github.io/graphql/draft/#sec-Errors ), this field is not always mandatory (only "If an error can be associated to a particular point in the requested GraphQL document", but there might be cases where it is not associated to a particular point, for example when the issue is global to the request method itself. And anyway, it says "it **should** contain an entry with the key locations", and not "it must" that's used for paths)

**Expected behavior**

Allow request with only a "message" key, or "message" + "path". This should be quite easy, but probably a breaking change for existing users of the library ?

HofmannZ:
Good catch, we want to align our package with the GraphQL spec. You PR is more then welcome!

PR #64, opened 2018-08-21
Error handling when location is not provided

adelcasse:
Fixes issue #60 by allowing an error response without location


HofmannZ:
Thanks for the fix! Please make sure all commits are GPG signed.

PR #78, opened 2018-09-03
FIX: runMutation return Future<void>

endigo:


HofmannZ:
Change the typedef to return void, because the function does never resolver with data.

PR #80, opened 2018-09-03
Only create a new observable query if options have changed

HofmannZ:
### Breaking changes

n/a

#### Fixes / Enhancements

- Only create a new observable query if options have changed. @HofmannZ

#### Docs

n/a

### DIFF
--- a/lib/src/core/query_options.dart
+++ b/lib/src/core/query_options.dart
@@ -137,6 +137,7 @@ class WatchQueryOptions extends BaseOptions {
     this.context,
   });
 
+  /// Checks if the [WatchQueryOptions] in this class are equal to some given options.
   bool areEqualTo(WatchQueryOptions otherOptions) {
     return !_areDifferentOptions(this, otherOptions);
   }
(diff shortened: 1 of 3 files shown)


lib/src/core/query_options.dart
```diff
@@ -136,4 +136,56 @@ class WatchQueryOptions extends BaseOptions {
...
+    if (a.pollInterval != b.pollInterval) {
+      return true;
+    }
+
+    if (a.fetchResults != b.fetchResults) {
+      return true;
+    }
+
+    // compare variables last, because maps take more time
+    return _areDifferentVariables(a.variables, b.variables);
```
eusdima: Nice optimization

lib/src/core/query_options.dart
```diff
@@ -136,4 +136,56 @@ class WatchQueryOptions extends BaseOptions {
...
+    Map b,
+  ) {
+    if (a.length != b.length) {
+      return true;
+    }
+
+    bool areDifferent = false;
+
+    a.forEach((key, value) {
+      if (b[key] != a[key] || (!b.containsKey(key))) {
```
eusdima: Compare values

HofmannZ: Nice catch

PR #85, opened 2018-09-04
Added optional parameter `initPayload`

lordgreg:
This PR adds the optional parameter to the static `connect` function. initPayload is a type of `Map<String, String>`, which will then be converted to JSON format. As specified by Apollo Client and required subscription authorisation, we could then simply do:

```
  Map<String, String> initPayload = new Map();
  initPayload['accessToken'] = Auth.accessToken;

    await SocketClient.connect(
      Static.BASE_URL_GRAPHQL_SUB,
      initPayload: initPayload
    )
    .then((createdClient) {
      client = createdClient;
      connections[operationName] = client;
    });
```

That being used, `connection_init` should now send the payload with `accessToken` to the websocket stream.

Resolves: https://github.com/zino-app/graphql-flutter/issues/74

### DIFF
--- a/lib/src/socket_client.dart
+++ b/lib/src/socket_client.dart
@@ -16,22 +16,26 @@ class SocketClient {
     final Map<String, String> headers = const {
       'content-type': 'application/json',
     },
+    final Map<String, String> initPayload
   }) async {
+    _initPayload = initPayload;
+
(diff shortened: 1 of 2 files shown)


HofmannZ:
Looks good to me, only not signed. Can you open a new PR with all commits signed?

Issue #95, opened 2018-09-11
How to decode UTF-8 chars in response?


endigo:
We have a Cyrillic (Mongolian) texts returned from GraphQL API.

But Cyrillic texts are obscured. Something like this.
![screen shot 2018-09-11 at 4 30 37 pm](https://user-images.githubusercontent.com/1501644/45348104-4cc3c400-b5e0-11e8-9015-60516c43f6c5.png)

We write a custom UTF-8 converter.

```dart
static String utf8convert(String text) {
    List<int> bytes = text.toString().codeUnits;
    return utf8.decode(bytes);
}
```

And the result is here
![screen shot 2018-09-11 at 4 35 19 pm](https://user-images.githubusercontent.com/1501644/45348247-b9d75980-b5e0-11e8-8a1c-9f5413e1a7cc.png)

HofmannZ:
Might be related #94

HofmannZ:
@endigo which version of the client are you using?

endigo:
@HofmannZ I'm using 1.0.0-alpha6.

HofmannZ:
@endigo Which `Content-Type` is your endpoint returning? Should be: `Content-Type: 'application/json; charset=utf-8'`, so that the http client can decode utf8.

endigo:
in the server's response header?

HofmannZ:
@endigo Yes, in the server's response header.

HofmannZ:
Fixed with 1.0.0-alpha.8

kieusonlam:
Hi guys,

Does the issue is fixed or just me still get this problem :( 

My package version: 1.0.0-alpha.9

![image](https://user-images.githubusercontent.com/13093313/46063129-18cad000-c196-11e8-9e90-cdcb227a583d.png)

AngeloContrera5:
@endigo thank u so much :3 <3

PR #101, opened 2018-09-14
Handle charset encoding in responses

kolja-esders:
#### Fixes / Enhancements

The charset encoding of the response is currently not taken into account when decoding the JSON body.

According to the JSON RFC spec, the body should be encoded as unicode with the default being `UTF-8`. In contrast to the specification, the body is currently decoded as `latin1` since this is the default encoding in the Dart library when there is no charset specified in the `Content-Type` header.

This PR should fix the issue by using `UTF-8` as the default encoding while also taking into account any charset specified in the header.

### DIFF
--- a/lib/src/link/http/link_http.dart
+++ b/lib/src/link/http/link_http.dart
@@ -3,6 +3,7 @@ import 'dart:convert';
 
 import 'package:meta/meta.dart';
 import 'package:http/http.dart';
+import 'package:http_parser/http_parser.dart';
 
 import 'package:graphql_flutter/src/link/link.dart';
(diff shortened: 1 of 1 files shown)


kolja-esders:
Should also point out that according to the RFC standard the charset encoding should not even be set in the `Content-Type` header since this was discussed in #95. Ideally, the encoding should be inferred based on the first 4 byte in the body.

This PR helps in so far as it uses `UTF-8` instead of `latin1` as default encoding and also covers cases where people decide to set a custom charset. Should be improved upon at a later stage by conforming to RFC.

PR #110, opened 2018-10-02
Fix Query variable change

### LINKED ISSUE #103: QueryState.didChangeDependencies not being called when variables change
**Describe the bug**
When I pass in a new QueryOptions object with different variables to a Query widget it sends a query with the variables from the previous build, not the new variables. I have traced the bug to the didChangeDependencies of QueryState not being called when the variables in the QueryOptions change (I haven't tested to see if its the same for other properties of the Query widget or QueryOptions object)

**To Reproduce**
create a widget that builds a Query widget that takes a QueryOptions that take in a variable map that has a single entry which is a string to another string which is a prop. render that widget multiple times in a row with different values for the prop that is being used as a variable to the Query. set a breakpoint on the first line of didChangeDependencies in the QueryState widget. It will run on the first mount but not on subsequent changes.

**Expected behavior**
When I pass in new variables to a Query widget it should send a query with the new variables, not the ones from the previous build

 - OS: macOS High Sierra 
 - Version 10.13.6

**Smartphone (please complete the following information):**
 - Device: Pixel XL (original)
 - OS: Android 9
 - Version Security Patch september 5, 2018

micimize:
closes #103
The method we needed was `didUpdateWidget`, which is triggered by prop changes.
Still left in a `didChangeDependencies` query, which is called whenever `client` changes, as well as after `initState` for the initial query.

Also:
* added a variable input to the example to test the change
* removed the `.replaceAll('\n', ' ')` from operations, as graphql allows for `# comments` that mangle documents if joined

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -54,11 +54,11 @@ class MyHomePage extends StatefulWidget {
 }
 
 class _MyHomePageState extends State<MyHomePage> {
-  int numberOfRepos = 50;
+  int nRepositories = 50;
 
   void changeQuery(String number) {
     setState(() {
-      numberOfRepos = int.parse(number) ?? 50;
+      nRepositories = int.parse(number) ?? 50;
     });
   }
 
@@ -69,7 +69,7 @@ class _MyHomePageState extends State<MyHomePage> {
         title: Text(widget.title),
       ),
       body: Container(
-        padding: const EdgeInsets.fromLTRB(8.0, 0.0, 8.0, 0.0),
+        padding: const EdgeInsets.symmetric(horizontal: 8.0),
         child: Column(
           mainAxisAlignment: MainAxisAlignment.start,
           mainAxisSize: MainAxisSize.max,
@@ -84,7 +84,7 @@ class _MyHomePageState extends State<MyHomePage> {
             Query(
               options: QueryOptions(
                 document: queries.readRepositories,
-                variables: <String, dynamic>{'numberOfRepos': numberOfRepos},
+                variables: <String, dynamic>{'nRepositories': nRepositories},
                 pollInterval: 4,
                 // you can optionally override some http options through the contexts
                 //
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -5,6 +5,8 @@ import 'package:graphql_flutter/graphql_flutter.dart';
 import './mutations/addStar.dart' as mutations;
 import './queries/readRepositories.dart' as queries;
 
+const String YOUR_PERSONAL_ACCESS_TOKEN = '<YOUR_PERSONAL_ACCESS_TOKEN_HERE>';
+
 void main() => runApp(MyApp());
 
 class MyApp extends StatelessWidget {
@@ -13,7 +15,7 @@ class MyApp extends StatelessWidget {
     final HttpLink link = HttpLink(
       uri: 'https://api.github.com/graphql',
       headers: <String, String>{
-        'Authorization': 'Bearer <YOUR_PERSONAL_ACCESS_TOKEN>',
+        'Authorization': 'Bearer $YOUR_PERSONAL_ACCESS_TOKEN',
       },
     );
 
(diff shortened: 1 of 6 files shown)


example/lib/main.dart
```diff
@@ -52,94 +54,126 @@ class MyHomePage extends StatefulWidget {
...
-                    title: Text(repository['name']),
-                    onTap: () {
-                      // optimistic ui updates are not implemented yet, therefore changes may take some time to show
-                      addStar(<String, dynamic>{
-                        'starrableId': repository['id'],
-                      });
-                    },
-                  );
+      body: Container(
+        padding: const EdgeInsets.fromLTRB(8.0, 0.0, 8.0, 0.0),
```
HofmannZ: Prefer `EdgeInsets.symmetric(horizontal: 8.0)` over `EdgeInsets.fromLTRB`.

example/lib/main.dart
```diff
@@ -52,94 +54,126 @@ class MyHomePage extends StatefulWidget {
 }
 
 class _MyHomePageState extends State<MyHomePage> {
+  int numberOfRepos = 50;
```
HofmannZ: Prefer `nRepositories` over `numberOfRepos`.

lib/src/widgets/query.dart
```diff
@@ -62,22 +44,39 @@ class QueryState extends State<Query> {
       fetchResults: true,
       context: widget.options.context,
     );
+  }
+
+  void query(WatchQueryOptions options) {
```
HofmannZ: Maybe name the function `_initQuery`, and include the `shouldCreateNewObservable` logic from the old `didChangeDependencies` method.

Currently changes to the client will trigger `didChangeDependencies` which calls `query` directly and does not close any open streams in the `observableQuery`.

lib/src/widgets/query.dart
```diff
@@ -62,22 +44,39 @@ class QueryState extends State<Query> {
...
-      }
+  @override
+  void didUpdateWidget(Query oldWidget) {
+    super.didUpdateWidget(oldWidget);
 
-      await observableQuery.close();
-    }
+    final WatchQueryOptions newOptions = _options;
+
+    final bool shouldCreateNewObservable =
```
HofmannZ: Prefer to use `!observableQuery.options.areEqualTo(_options);` inside the if statement over two separate variable declarations.

lib/src/core/query_options.dart
```diff
@@ -154,11 +154,7 @@ class WatchQueryOptions extends QueryOptions {
       return false;
     }
 
-    if (a == null && b != null) {
-      return true;
-    }
-
-    if (a != null && b == null) {
+    if (a == null || b == null) {
```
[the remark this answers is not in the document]
micimize: the `(a == null && b == null)` clause above it handles both nulls, so here we know it's either one or the other

lib/src/widgets/query.dart
micimize: good catch - added `observableQuery?.close()` to `_initQuery`. I left the `shouldCreateNewObservable` logic in `didUpdateWidget` because I figure if, for instance, the entire client changes, we should probably rebuild the query

lib/src/core/query_options.dart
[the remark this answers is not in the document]
HofmannZ: @micimize that is totally correct.

HofmannZ:
@micimize Looks good! Please make sure all commits are GPG signed.

micimize:
@HofmannZ great! Rebased, but had to edit the commit messages and force push for @gpg to pick up the signatures I added

micimize:
Using this from our depending project, I've discovered hot updates trigger `didUpdateWidget`, but not `didChangeDependencies`. In particular:
* updating cache -> updates query results
* hot reload -> cache is still correct, query results reset to stale data

Removing the `(!observableQuery.options.areEqualTo(_options))` constraint solves this problem, however I'm a bit uncomfortable with how little I understand what lifecycle methods fire when

HofmannZ:
@micimize Maybe we can check if hot update has triggered, and reset the stream? Merging for now, because it closes #103

PR #111, opened 2018-10-04
NormalizedInMemoryCache

### LINKED ISSUE #109: object-level / flattened caching
We're currently doing [operation level caching](https://github.com/zino-app/graphql-flutter/blob/next/lib/src/core/query_manager.dart#L124) instead of the [object-level caching](https://www.apollographql.com/docs/react/advanced/caching.html) available in apollo. This means disparate operations on the same resource result in inconsistent cache data.

The first steps here look to me to be:
* add an optional `dataIdFromObject` equivalent
* traverse results and cache them based on the `dataId`
* replace values in operation cache with references

This would allow mutation results to automatically invalidate/update query results, which would also be a good step towards user friendly optimistic updates.

**Additional context**
While eventually it'd be nice to [inject `__typename` automatically](https://github.com/apollographql/apollo-client/blob/c3561e73957799bbf0d7f87c4a3ebcfee6f40e3a/packages/apollo-utilities/src/transform.ts#L168) the way apollo does, maybe using angel's [graphql_parser](https://github.com/angel-dart/graphql/tree/master/graphql_parser), I think we can punt that to userland initially.

micimize:
closes #109

#### Fixes / Enhancements
- Added `lib/src/cache/normalized_in_memory.dart` that overrides `InMemoryCache`'s `read` and `write` functionality as described in #109
- Added some tests for it, runnable with `flutter test test/normalized_in_memory_test.dart`

#### Docs
- Added a [normalization subsection](https://github.com/micimize/graphql-flutter/tree/normalization#normalization) to the cache docs

I deferred adding `__typename` injection and making `typenameDataIdFromObject` a default mainly because `graphql_parser` doesn't have [stringification support](https://github.com/angel-dart/graphql/issues/10) yet, but will likely PR angel for it soon.


HofmannZ:
@micimize Thanks for the PR! Why did you choose not to use `typenameDataIdFromObject` as a default?

Also I think this is the preferred caching behaviour, and it should replace the default `InMemoryCache`.

micimize:
@HofmannZ Didn't want to imply our normalization is somehow automatic. Also seemed more consistent, given that adding `__typename` to every field still needs to be explicit.

PR #117, opened 2018-10-11
🔧 fixed bug when set state was called after widget was unmounted

SirKuryaki:
#### Fixes / Enhancements

- Fixed: Query Widget was calling set state after being unmounted ([https://docs.flutter.io/flutter/widgets/State/mounted.html]).


HofmannZ:
PR looks good. Make sure you sign all your commits

PR #119, opened 2018-10-11
Subscriptions

### LINKED ISSUE #102: socketClient is always null
**Description of the bug**
`socketClient` is initialized but never set to the actual connection ([reference](https://github.com/zino-app/graphql-flutter/blob/a34fea21830b56b4feb37a19c46b08d1873544c2/lib/src/socket_client.dart#L8)). 
This also results in an error when using the subscription-Widget, because [`socketClient.connect()`](https://github.com/zino-app/graphql-flutter/blob/a34fea21830b56b4feb37a19c46b08d1873544c2/lib/src/widgets/subscription.dart#L49) is called on null.

**Reproduction**
n/a

**Expected behavior**
There should be a way to set the [`SocketClient`](https://github.com/zino-app/graphql-flutter/blob/a34fea21830b56b4feb37a19c46b08d1873544c2/lib/src/socket_client.dart#L8) to the websocket connection reference.

**Additional info**
It would also be a good idea to update the readme file.

Assuming I didn't miss a thing are the proposals for an alternative implementation?

jpJuni0r:
Implement Subscriptions Widget correctly.

### Breaking changes

- Subscription Widget now requires a socketClient parameter

#### Fixes / Enhancements

- fixed: require socketClient parameter on Subscription Widget

#### Docs

n/a

#### Notes

Fixes #102

### DIFF
--- a/lib/src/socket_client.dart
+++ b/lib/src/socket_client.dart
@@ -5,8 +5,6 @@ import 'package:uuid/uuid.dart';
 
 import 'package:graphql_flutter/graphql_flutter.dart';
 
-SocketClient socketClient;
-
 class SocketClient {
   final Uuid _uuid = Uuid();
   final GraphQLSocket _socket;
--- a/lib/src/widgets/subscription.dart
+++ b/lib/src/widgets/subscription.dart
@@ -16,6 +16,7 @@ typedef SubscriptionBuilder = Widget Function({
 class Subscription extends StatefulWidget {
   final String operationName;
   final String query;
+  final SocketClient socketClient;
   final dynamic variables;
   final SubscriptionBuilder builder;
   final OnSubscriptionCompleted onCompleted;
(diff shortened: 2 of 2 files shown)


HofmannZ:
Thanks for the PR! But injecting the socket client into the widget is not the way we think we should implement this. Instead we think we should merge the socket client into the already exciting client.

We already made some steps for it by handeling all executions as streams that perform a certain operation. Just need to refactor the client a bit and create a websocket link so it handles sockets correctly.

jpJuni0r:
Sure, I agree with that. Is there anything I can be helpful with?

So far I can't pull the websocket url from the client because only accepts an http endpoint.

HofmannZ:
@jpJuni0r Yeah, we've already made some steps allowing the merger of the two clients, in the new client. Bu have not made any progress yet on the subscriptions client side. That could be something you could help us with.

PR #120, opened 2018-10-11
Fix mutation calling onCompleted for loading state

rafaelring:
After the introduction of the initial loading state to false on PR #112, the first emission of the observableQuery observable is a query result with the `loading: true`. 

The unexpected side effect of this change is that the `onCompleted` callback, which listens only to the first emission, would be called only for the loading state change and not for the query response itself.

This PR changes the `onCompleted` subscription to only call the method and cancel the subscription after the loading changes back to false.


knewter:
:+1: I came here to submit this PR :)

HofmannZ:
Nice catch! PR looks good to me

PR #137, opened 2018-11-20
Bugfix/subscription init payload

lordgreg:

### Breaking changes

Init payload was `json.encoded` which, in last iterations of alpha version of GraphQL, breaks subscriptions. Checking websocket connections, init payload has too many backslashes added to the quotes.

#### Fixes / Enhancements

Removed encoding from init `toJson` function

### DIFF
--- a/lib/src/websocket/messages.dart
+++ b/lib/src/websocket/messages.dart
@@ -57,10 +57,10 @@ class InitOperation extends GraphQLSocketMessage {
     jsonMap['type'] = type;
 
     if (payload != null) {
-      jsonMap['payload'] = json.encode(payload);
+      jsonMap['payload'] = payload;
     }
 
-    return json.encode(jsonMap);
+    return jsonMap;
   }
 }
 


lordgreg:
@HofmannZ - MR has now all signed commits and can be reviewed. Please let me know if there are any open questions.

lib/src/websocket/messages.dart
```diff
@@ -57,10 +57,10 @@ class InitOperation extends GraphQLSocketMessage {
     jsonMap['type'] = type;
 
     if (payload != null) {
-      jsonMap['payload'] = json.encode(payload);
+      jsonMap['payload'] = payload;
     }
 
-    return json.encode(jsonMap);
+    return jsonMap;
```
HofmannZ: This part should still be encoded, line 60 was cousing the issue I believe

ChristopherKlammt: For my code, it is necessary to return the plain jsonMap instead of encoding it in order for the init payload to work properly.

eusdima: Wouldn't it be possible to add an unescaped method that doesn't do the encoding alongside the regular one?

And one more thing, you can't justify a pull by saying that it fixes your use case.

micimize:
from #242
> `initPayload` should definitely be at least as flexible as `Map<String, dynamic>`. Will follow through in #137 to  see exactly why it is so constrained

@eusdima why is this controversial? I think `initPayload` should probably be a generic entry point that's flexible enough to handle whatever initialization structure the particular server needs, which means it should really accept any valid JSON structure

micimize:
closing in favor of #277

Issue #144, opened 2018-12-31
Implement fetchMore API for pagination


chimon2000:
**Is your feature request related to a problem? Please describe.**
Currently, there is not an elegant way to do pagination with a GraphQL query.  Apollo supports this thru the fetchMore API that is part of the query result.

**Describe the solution you'd like**
In Apollo, the fetchMore function allows you to do a new GraphQL query and merge the result into the original result.  You can override variables, and provide a callback function to update the previous query.  

Here's a simple example in in JavaScript

```js
fetchMore({
  variables: {
    offset: data.feed.length
  },
  updateQuery: (prev, { fetchMoreResult }) => {
    if (!fetchMoreResult) return prev;
    return Object.assign({}, prev, {
      feed: [...prev.feed, ...fetchMoreResult.feed]
    });
  }
});
```

**Describe alternatives you've considered**
Right now the best solution I can come up with is to implement this logic with a GraphQLConsumer that gives me access to the GraphQLClient. This isn't the worst, but it does require quite a bit of boilerplate code.

**Additional context**
https://www.apollographql.com/docs/react/features/pagination.html

wwwdata:
Hi, we are currently evaluating this plugin and I also asked myself how to implement `fetchMore`. Actually it is quite simple to make a widget that extends the standard `Query` widget and provides this functionality like in the javascript apollo client. I have just extended the `builder` with the `fetchMore` callback.

I could also open a pull request into the `next` branch with this. But the question would be: Make this its own widget or integrate it into the base `Query` widget? For that we would need to permanently add the `fetchMore` parameter (maybe as optional one) into the `builder` function signature. 

```dart
import 'package:flutter/material.dart';
import 'package:graphql_flutter/graphql_flutter.dart';

typedef void FetchMoreCallback(
  Map<String, dynamic> variables,
  MergeResults mergeResults,
);

typedef Map<String, dynamic> MergeResults(
  dynamic prev,
  dynamic moreResults,
);

typedef Widget FetchMoreBuilder(
  QueryResult result,
  FetchMoreCallback fetchMore,
);

class QueryFetchMore extends StatefulWidget {
  final QueryOptions options;
  final FetchMoreBuilder builder;

  const QueryFetchMore({Key key, this.options, this.builder}) : super(key: key);

  @override
  QueryFetchMoreState createState() {
    return new QueryFetchMoreState();
  }
}

class QueryFetchMoreState extends State<QueryFetchMore> {
  bool _init = false;
[...]

PR #146, opened 2019-01-02
Add refetch to QueryResult

lohmander:
Adding a refetch function to QueryResult. Perhaps there's another better way that also can be easily implemented, but this seems to do the job for now.

### Breaking changes

No breaking changes.

#### Fixes / Enhancements

- Added `refetch` to QueryResult

#### Docs

None (yet).

### DIFF
--- a/lib/src/core/observable_query.dart
+++ b/lib/src/core/observable_query.dart
@@ -28,7 +28,10 @@ class ObservableQuery {
     );
   }
 
-  Stream<QueryResult> get stream => controller.stream;
+  Stream<QueryResult> get stream => controller.stream.map((QueryResult result) {
+        result.refetch = fetchResults;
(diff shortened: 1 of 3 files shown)


venkatd:
@lohmander we need this as well! Is there anything remaining to be done here?

Happy to help get this merged however I can.

olup:
Just a note, maybe api for `refetch` and `fetchMore` should be provided in a coherent way ? See #144

lohmander:
@venkatd I've not had time to familiarize myself with the codebase as a whole, so this solution might not be inline with that the package authors have in mind. But it does the job and we already use the fork in a production app.

HofmannZ:
Closing this PR for now and continuing discovery on issue #144.

PR #163, opened 2019-02-03
WebSocket functionality with automatic reconnection and keep alive message handling

szantogab:
This pull request brings advanced WebSocket functionality to the library. It implements the `WebsocketLink` class. Also, it adds automatic Websocket reconnection, keep alive message handling, and the ability to send not only subscriptions but also query and mutation over the `WebsocketLink`. All parameters can be configured via the `SocketClientConfig` class.

Basically, users just instantiate `SocketClient` class with the given config, and it takes care of the rest:

```
    final WebSocketLink websocketLink = WebSocketLink(SocketClient(
      'ws://api.github.com/graphql',
      headers: <String, String>{
        'Authorization': 'Bearer $YOUR_PERSONAL_ACCESS_TOKEN',
      },
config: SocketClientConfig(autoReconnect: true, inactivityTimeout: Duration(seconds: 15), delayBetweenReconnectionAttempts: Duration(seconds: 5), queryAndMutationTimeout: Duration(seconds: 10)),
    ));
```

A `SocketClient` instance can now de disposed with the `dispose()` method.

A stream is available to users to listen for connection changes:
```
connectionState.listen((SocketConnectionState state) {
      print("Websocket state changed to: $state");
    });
```

### Breaking changes
Removed `GraphQLSocket` class.

Hope you like it.


szantogab:
@HofmannZ any chance to review this PR in the near future?

HofmannZ:
Amazing PR. Can you resolve the merge issues?

example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -8,7 +8,6 @@
 
 /* Begin PBXBuildFile section */
 		1498D2341E8E89220040F4C2 /* GeneratedPluginRegistrant.m in Sources */ = {isa = PBXBuildFile; fileRef = 1498D2331E8E89220040F4C2 /* GeneratedPluginRegistrant.m */; };
-		2D5378261FAA1A9400D5DBA9 /* flutter_assets in Resources */ = {isa = PBXBuildFile; fileRef = 2D5378251FAA1A9400D5DBA9 /* flutter_assets */; };
```
HofmannZ: Why did you take this out?

example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -42,7 +41,6 @@
 /* Begin PBXFileReference section */
 		1498D2321E8E86230040F4C2 /* GeneratedPluginRegistrant.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = GeneratedPluginRegistrant.h; sourceTree = "<group>"; };
 		1498D2331E8E89220040F4C2 /* GeneratedPluginRegistrant.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = GeneratedPluginRegistrant.m; sourceTree = "<group>"; };
-		2D5378251FAA1A9400D5DBA9 /* flutter_assets */ = {isa = PBXFileReference; lastKnownFileType = folder; name = flutter_assets; path = Flutter/flutter_assets; sourceTree = SOURCE_ROOT; };
```
HofmannZ: Why did you take this out?

example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -85,7 +83,6 @@
 		9740EEB11CF90186004384FC /* Flutter */ = {
 			isa = PBXGroup;
 			children = (
-				2D5378251FAA1A9400D5DBA9 /* flutter_assets */,
```
HofmannZ: Why did you take this out?

example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -213,7 +210,6 @@
 				3B3967161E833CAA004F5970 /* AppFrameworkInfo.plist in Resources */,
 				9740EEB41CF90195004384FC /* Debug.xcconfig in Resources */,
 				97C146FE1CF9000F007C117D /* Assets.xcassets in Resources */,
-				2D5378261FAA1A9400D5DBA9 /* flutter_assets in Resources */,
```
HofmannZ: Why did you take this out?

szantogab:
@HofmannZ can you please review this PR again? I've merged the next branch into this branch.

filippofilip95:
I've just tested your implementation and it Works great :)

but one point, @szantogab  client is reconnecting after connection is lost in specified interval `delayBetweenReconnectionAttempts` thats fine. But Iam not able to resubscribe to all my subsciptions. 

It would be great to have some callback fe. `onReconnect` which will be called after connection is established again. 

Do you think is goo idea to implement it ? or any idea hot to solve resubscription ? 

Thanks.

szantogab:
@filippofilip95 there is a stream called `connectionState`, which you can use to do anything you want when the connection state changes.

HofmannZ:
Shouldn't we just concatenate the two streams, and have the subscription stream filter out Query and Mutation operations.

This way we could get the socket client from the context and we no longer have to provide it to the subscription widget.

What are your thoughts?

szantogab:
Getting the socket client from the `context ` is a great idea!

HofmannZ:
@szantogab I mean that the socket link would filter the operation based on the operation type end then eighter subscribe to the socket client or forward the operation the next link (ex. the HTTP link)

This way you can completely abstract the socket client away for the widget layer. Also, it would allow you to use the same auth link for both the socket link and the HTTP link.

filippofilip95:
@szantogab  Thanks your reply, you helped me a lot. But one more thing. How can i close a listening to stream without disposing a SocketClient.

Iam using it like this.

```
 _wsClient = SocketClient( 'ws://192.168.1.14:4000');

 _stream = _wsClient.subscribe(
          SubscriptionRequest(
            _operationName,
            updatedPlayBackQueueSubscription,
            variables,
          ),
          true,
        );

        _stream.listen(_onStreamData);

```

And the thing is that when i subscribe again like this `_stream = _wsClient.subscribe(...)`  method _onStreamData is firing for previous and actual subscription Is it possible to close previous subscribtion ?

is it only way to do it by `_wsClient.dispose()` ?

Iam note very handy with Streams .

example/lib/main.dart
```diff
@@ -9,6 +9,8 @@ const String YOUR_PERSONAL_ACCESS_TOKEN = '<YOUR_PERSONAL_ACCESS_TOKEN>';
 void main() => runApp(MyApp());
 
 class MyApp extends StatelessWidget {
+  MyApp();
+
```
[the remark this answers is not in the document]
szantogab: It not necessary, we can remove it.

szantogab:
@HofmannZ Sounds great. Should I give it a go to implement it?

szantogab:
@HofmannZ If I understand correctly, your solution would rule out the possibility of sending query and mutation operation via websocket. I currently have an app where all communication occurs via websocket, and we don't have an HTTP transport, so I think this way we would lose the flexibility we currently have regarding transport.

szantogab:
@HofmannZ @filippofilip95 As @HofmannZ pointed out, the `SocketClient` should be abstracted away from the `Subscription` widget.

I think I found a solution for that with this new commit.

Basically what I did is that each `Link` decides whether it can handle the given `Operation` or not. For example, the `HttpLink` class cannot handle subscriptions, so it forwards the request to the next link, or throws an exception if no forwardLink is passed to it.

So we achieve full abstraction with a Link created like this:
`final Link link = authLink.concat(httpLink).concat(websocketLink);`

Any query, mutation, or subscription will go through this chain of `Link`s in the given order.
So with this example this is what happens:

1. The `AuthLink` obtains the headers and put them in the context.
2.  The `HttpLink` checks whether the current operation is supported. If it's a query or mutation it simply executes the request. However if it's a subscription, it passes the `Operation` to the next `Link`.
3. The `WebSocketLink` supports all types of operations, so it will execute it. 

Also, this way if we want, we can have the `WebSocketLink` handle everything without any other `Link`s.


Moreover, the `HttpLink` and `WebSocketLink` both respects the `AuthLink`. The `WebSocketLink` implementation was a bit tricky in this regard, because by changing the header, the socket connection needs to be re-established to be able to send the new headers. There's a `reconnectOnHeaderChange` flag that is enabled by default to support such behaviour.

  Please see `WebSocketLink` dartdoc for more details.


I hope you like it, any idea, review, or critics are welcome!

HofmannZ:
@szantogab That is exactly what I meant. I will try to get some time in to review it later this day.

example/ios/Runner.xcodeproj/project.pbxproj
micimize: I think I saw this in other PRs as well and thought it was just a change in flutter build output. made #206 to investigate and follow up in

szantogab:
@micimize I developed the server based on Spring Boot and graphql-spqr library.

filippofilip95:
@micimize I'am using [Prisma](https://www.prisma.io/) with [graphql-yoga](https://github.com/prisma/graphql-yoga). But will migrate yoga to [ApolloServer](https://www.apollographql.com/docs/apollo-server/). Bot of them with have support for subscriptions.

PR #164, opened 2019-02-06
Fix erroneous closing of observableQuery

knewter:
The previous code closed the `observableQuery` any time
`didChangeDependencies` was called, whether or not it was creating a new
one. This meant you could have a closed stream that events were added
to, which caused the following error:

```
flutter: Another exception was thrown: Bad state: Cannot add new events
after calling close
```

### Breaking changes

- None

#### Fixes / Enhancements

- Fixed calling `runMutation` multiple times on the same screen in the case where a rebuild was triggered on `Mutation` but no dependencies that it cared about were changed.
  - The error was `Bad state: Cannot add new events after calling close`

### DIFF
--- a/lib/src/widgets/mutation.dart
+++ b/lib/src/widgets/mutation.dart
@@ -95,11 +95,10 @@ class MutationState extends State<Mutation> {
       if (observableQuery.options.areEqualTo(options)) {
         shouldCreateNewObservable = false;
       }
-
-      observableQuery.close();
     }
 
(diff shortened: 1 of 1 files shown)


micimize:
I also ran into this in #199 - it has the equivalent fix as well as some `QueryLifecycle` management that should help with some similar cases. Thanks for the contribution though! I'll ping you when #199 is merged and released so you can test and make sure it fixes this issue

PR #171, opened 2019-02-16
✨Added the Auth Link

HofmannZ:
This PR introduces the brand new Auth Link!

### Breaking changes

N/A

#### Fixes / Enhancements

- Added Auth Link class.

#### Docs

- Updated the example to use the new AuthLink.

### DIFF
--- a/lib/src/link/auth/link_auth.dart
+++ b/lib/src/link/auth/link_auth.dart
@@ -1,0 +1,38 @@
+import 'dart:async';
+
+import 'package:graphql_flutter/src/link/link.dart';
+import 'package:graphql_flutter/src/link/operation.dart';
+import 'package:graphql_flutter/src/link/fetch_result.dart';
+
+typedef GetToken = String Function();
+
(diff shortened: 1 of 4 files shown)


lib/src/link/auth/link_auth.dart
```diff
@@ -0,0 +1,38 @@
...
+class AuthLink extends Link {
+  AuthLink({
+    this.getToken,
+  }) : super(
+          request: (Operation operation, [NextLink forward]) {
+            StreamController<FetchResult> controller;
+
+            Future<void> onListen() async {
+              try {
+                final String token = getToken();
```
michael-golfi: Just checking in - is there a way here to also support async token calls?

Particularly in my use case (maybe others as well), I need to support cases where the token gets expired and needs to be refreshed and that would require doing the token call asynchronously.

micimize: I think #178 should solve this, if not go ahead and open a new issue

PR #174, opened 2019-02-19
Fix cache updating subqueries that does not have typename

degroote22:
It is a follow up to this PR: https://github.com/zino-app/graphql-flutter/pull/156
I ended up still having the issue where the subquery would override the superquery when there wasn't a typename on the field.

This PR adds tests for this edge case and fixes it.

I'm sorry I didn't put this all on the same PR.

### DIFF
--- a/lib/src/cache/in_memory.dart
+++ b/lib/src/cache/in_memory.dart
@@ -2,10 +2,24 @@ import 'dart:async';
 import 'dart:collection';
 import 'dart:convert';
 import 'dart:io';
-
 import 'package:path_provider/path_provider.dart';
-
 import 'package:graphql_flutter/src/cache/cache.dart';
+
+void _recursiveAddAll(
+  Map<String, dynamic> oldData,
+  Map<String, dynamic> newData,
+) {
+  newData.forEach((String key, dynamic value) {
+    if (oldData.containsKey(key) &&
+        oldData[key] is Map &&
+        value != null &&
+        value is Map) {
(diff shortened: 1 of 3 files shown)


lib/src/cache/in_memory.dart
```diff
@@ -2,11 +2,25 @@ import 'dart:async';
 import 'dart:collection';
 import 'dart:convert';
 import 'dart:io';
-
 import 'package:path_provider/path_provider.dart';
-
 import 'package:graphql_flutter/src/cache/cache.dart';
 
+void _recursiveAddAll(
```
HofmannZ: Shouldn't this be a helper function inside `utilities/helpers.dart` ?

lib/src/cache/in_memory.dart
```diff
@@ -2,11 +2,25 @@ import 'dart:async';
 import 'dart:collection';
 import 'dart:convert';
 import 'dart:io';
-
```
HofmannZ: We tend to keep Dart packages separate from installed packages.

lib/src/cache/in_memory.dart
```diff
@@ -2,11 +2,25 @@ import 'dart:async';
...
 import 'package:path_provider/path_provider.dart';
-
```
HofmannZ: We also tend to keep project packages separate from dart and installed packages.

micimize:
@HofmannZ moved the helper and broke up the imports, so this is good to go.

PR #175, opened 2019-02-21
Fixes incomplete cache restoration issue

degroote22:
It follows https://github.com/zino-app/graphql-flutter/issues/152 and https://github.com/zino-app/graphql-flutter/pull/156  

__This basically fixes a race condition. `tl;dr` we can't read and write at the same time__:
On android the app will fire `inactive` and then `restored` when it is restored.
Then, while this is reading, it may read incomplete data because the writer may be writing data.
If it is writing, the file was overwritten anyway, so just hope that the data we have in memory is the correct one.

<details>
 <summary><b>extended context provided by <code>@degroote22</code></b></summary>

 I'm having a hard time creating a PR without the[ commits from Feb 19](https://github.com/zino-app/graphql-flutter/pull/174) which are unrelated to this issue but on the same file.[ This is the relevant commit.](https://github.com/zino-app/graphql-flutter/commit/54e8b65cde51a5a84791856cc449b2f1f9f0b1a1) The other ones are awaiting to be merged in another PR.

The reasoning behind the changes are written as comments in code.
I only have an Android device to test and this PR fixes my issues.
Like I said on the code, I'm in doubt about calling save when the app is inactive...
Is the reasoning that the user may have the app in this state and not paused and then the phone may turn off and the data will be saved?

This requires testing and I only have one device to test it.
Anyway, I have just one phone running Android 6 to test this and it works for me 🤷‍♂️ 
</details>

### DIFF
--- a/lib/src/cache/in_memory.dart
+++ b/lib/src/cache/in_memory.dart
@@ -4,7 +4,6 @@ import 'dart:convert';
 import 'dart:io';
 
 import 'package:path_provider/path_provider.dart';
-
 import 'package:graphql_flutter/src/cache/cache.dart';
 import 'package:graphql_flutter/src/utilities/helpers.dart'
     show deeplyMergeLeft;
@@ -38,11 +37,11 @@ class InMemoryCache implements Cache {
     if (_inMemoryCache.containsKey(key) &&
         _inMemoryCache[key] is Map &&
         value != null &&
-        value is Map) {
+        value is Map<String, dynamic>) {
       // Avoid overriding a superset with a subset of a field (#155)
       // this means deletions must be done by explicitly returning a field as null
       _inMemoryCache[key] = deeplyMergeLeft(<Map<String, dynamic>>[
-        _inMemoryCache[key],
+        _inMemoryCache[key] as Map<String, dynamic>,
         value,
       ]);
     } else {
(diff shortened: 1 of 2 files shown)


lib/src/cache/in_memory.dart
```diff
@@ -104,6 +122,18 @@ class InMemoryCache implements Cache {
...
 
   Future<HashMap<String, dynamic>> _readFromStorage() async {
+    if (_writingToStorage) {
+      // On android the app will fire 'inactive' and then 'restored'
+      // when it is restored. Then, while this is reading it may read
+      // incomplete data because the writer may be writing data.
+      // And if it is writing the file was overwritten anyway, so we don't
+      // have what to read and best hope that the data we have in memory
+      // is the correct one.
+      // TODO: should we call save when the app enters inactive (why?)
```
HofmannZ: Can you move this into the PR description?

micimize:
@degroote22 I moved your todo to where we actually call `inactive` on save. Am not sure I understand the fully, but please open a new issue for further discussion if you'd like to.

micimize:
@degroote22 I moved your todo to where we actually call `inactive` on save. Am not sure I understand the fully, but please open a new issue for further discussion if you'd like to.

smkhalsa:
I am still experiencing this behavior. On Android, when I turn on airplane mode then kill the app and reopen (and continue killing and reopening the app without turning data back on), there is data about 50% of the time.

PR #184, opened 2019-03-09
All the HTTP responses are now correctly decoded to utf8 before retur…

joanpablo:
…n from the Client.query() method

In an HTTP request to a server that returns a response containing characters like: "Jiménez" the 'graphql_flutter' library returns from the Client.query() method a string like: "Jim@nez" because the response wasn't decoded to UTF-8. This pull request is a 'hotfix' to this issue.

#### Fixes / Enhancements

- Fixed the json utf8 decoding in responses from Client.query()

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -47,7 +47,7 @@ class Client {
 
   Map<String, String> get headers => {
         'Authorization': 'Bearer $apiToken',
-        'Content-Type': 'application/json',
+        'Content-Type': 'application/json; charset=utf-8',
       };
 
   InMemoryCache get cache => this._cache;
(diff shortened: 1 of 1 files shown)


joanpablo:
Sorry I create the pull request over an old branch master I will check that error in the 1.0.0 beta1 and I will upload again the request if it is necessary

joanpablo:
I'm a little confuse, I make the hotfix in the master but it doesn't have the break changes of the previous tags like 1.0.0-alpa.11

joanpablo:
obviously fixed in branch 'next' 

final Encoding encoding = _determineEncodingFromResponse(response);
  final dynamic decodedBody = encoding.decode(response.bodyBytes);

PR #187, opened 2019-03-12
expose runQuery Method to trigger Query Manually

echonox:

### Breaking changes
This changes breaks the basic query api, now it adds an `runQuery` extra param in the builder list of parameters

#### Fixes / Enhancements

- Adds the ability to trigger the query manually, until now it would require to call the client manually, or use a poll interval to update the refetch the query

#### Docs

- Before:

```dart
...

Query(
  readRepositories, // this is the query you just created
  variables: {
    'nRepositories': 50,
  },
  pollInterval: 10, // optional
  builder: ({
    bool loading,
    var data,
    Exception error
  }) {
    if (error != null) {
      return Text(error.toString());
    }

    if (loading) {
      return Text('Loading');
    }

    // it can be either Map or List
    List repositories = data['viewer']['repositories']['nodes'];

    return ListView.builder(
      itemCount: repositories.length,
      itemBuilder: (context, index) {
        final repository = repositories[index];

        return Text(repository['name']);
    });
  },
);

...
```

After:

```dart
...

Query(
  readRepositories, // this is the query you just created
  variables: {
    'nRepositories': 50,
  },
  pollInterval: 10, // optional
  builder: ({
    bool loading,
    var data,
    Exception error,
    runQuery
  }) {
    if (error != null) {
      return Text(error.toString());
    }

    if (loading) {
      return Text('Loading');
    }

    // it can be either Map or List
    List repositories = data['viewer']['repositories']['nodes'];

    return ListView.builder(
      itemCount: repositories.length,
      itemBuilder: (context, index) {
        final repository = repositories[index];

        return Text(repository['name']);
    });
  },
);

...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,7 @@ Query(
     bool loading,
     var data,
     Exception error,
+    runQuery
   }) {
     if (error != null) {
       return Text(error.toString());
--- a/lib/src/widgets/query.dart
+++ b/lib/src/widgets/query.dart
@@ -4,10 +4,12 @@ import 'package:flutter/widgets.dart';
 import 'package:graphql_flutter/src/client.dart';
 import 'package:graphql_flutter/src/widgets/graphql_provider.dart';
 
+typedef void GetQueryResult();
 typedef Widget QueryBuilder({
   @required bool loading,
   Map<String, dynamic> data,
   Exception error,
+  GetQueryResult runQuery
 });
 
 class Query extends StatefulWidget {
@@ -160,6 +162,7 @@ class QueryState extends State<Query> {
       loading: loading,
       error: error,
       data: data,
+      runQuery: getQueryResult
     );
   }
 }


juicycleff:
I think refetch would be a better method to expose based on the apollo client design. I am currently implementing this future in my fork

echonox:
@juicycleff agree on that, shouldn't we have a strict naming for the api? as some methods aren't following the apollo client design, also I see your fork is way ahead of the main repo, and it seems you are using streams (bloc pattern), should we use your fork in production or stick with the main?

juicycleff:
To be honest, I just felt the revision process takes a long time here, and I've needed this library as mature as soon as possible, so my fork is a little bit different and merging back here will be a little harder than anticipated. I currently use my fork in production and is still WIP, however, I duplicated my fork to my company Org so as to get more contributors as it's a little bit organized differently from this. You can use mine if you wish, I published with flutter_graphql, but there's still a lot to be done.

juicycleff:
The project lives here now https://github.com/snowballdigital/flutter-graphql

micimize:
@echonox This is good, but active development is on `next`. I've gone ahead and implemented `refetch` on a fork from there and will be PRing soon 👍 

On `next`, `runMutation` is positional, and thus closer to the original apollo API.

PR #192, opened 2019-03-21
Verify errors are not null before JSON transform

mwalkerwells:
The jsonResponse is a list of nullable errors & can throw if assumed to be non-nullable...

**link_http.dart** (inspector)
```
> jsonResponse
Map (2 items)
0:"errors" -> List (1 item)
key:"errors"
value:List (1 item)
[0]:null
1:"data" -> Map (1 item)
key:"data"
value:Map (1 item)
0:"viewer" -> null
```

### DIFF
--- a/lib/src/core/query_manager.dart
+++ b/lib/src/core/query_manager.dart
@@ -192,9 +192,11 @@ class QueryManager {
     List<GraphQLError> errors;
 
     if (fetchResult.errors != null) {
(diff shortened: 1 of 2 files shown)


micimize:
The code's changed some structurally and this change is minor, so I added the check as `jsonResponse['errors']?.where(notNull)?.toList()` to #199 and am closing this

Issue #177, opened 2019-02-22
Exceptions are swallowed


venkatd:
Related to #140 

In most of the `try ... catch` clauses in the code, we are catching all exceptions. This results in errors unrelated to the library (such as network errors) getting swallowed which makes things very difficult to debug.

For example:
https://github.com/zino-app/graphql-flutter/blob/f7d50caf4e4999761ac8c128097ff90b3e19bb99/lib/src/widgets/query.dart#L90

https://github.com/zino-app/graphql-flutter/blob/f7d50caf4e4999761ac8c128097ff90b3e19bb99/lib/src/widgets/query.dart#L107

The solution would be to catch on specific exceptions. We do this in our own code for example:

```
  _login(BuildContext context) async {
    try {
      await auth.signIn();

      Navigator.of(context).pushReplacement(
        new MaterialPageRoute(
          builder: (context) {
            return DashboardContainer();
          },
        ),
      );
    } on GoogleSignInFailedError catch (error) {
      _showLoginFailed(context, error);
    } on TurtleSignInFailedError catch (error) {
      _showLoginFailed(context, error);
    }
  }
```

If there is any other exception besides what we explicitly catch, it will properly get raised so a developer can debug easily.

HofmannZ:
I think we could solve this by creating our own kind of exceptions so that we can handle them cleaner internally.

truongsinh:
closed by #193

PR #193, opened 2019-03-21
Better handling of cases where dart errors are throw during response …

mwalkerwells:
Not all errors thrown here are GraphQL errors and we should not show an error related to being unable to access `message`.

When I encountered this, it led to a lot of misdirection and debugging...


venkatd:
My thinking is it might be cleaner to create graphql-specific exception classes and only catch those. This way non-graphql errors propagate up as expected. I created an issue related to this here #177

PR #195, opened 2019-03-22
use stricter analysis style

truongsinh:

#### Enhancements
- Use stricter analysis style to elaborate whenever we cast


lib/src/cache/normalized_in_memory.dart
```diff
@@ -2,7 +2,7 @@ import 'package:meta/meta.dart';
 import 'package:graphql_flutter/src/utilities/traverse.dart';
 import 'package:graphql_flutter/src/cache/in_memory.dart';
 
-typedef String DataIdFromObject(Object node);
+typedef DataIdFromObject = String Function(Object node);
```
micimize: I personally prefer the `typedef String DataIdFromObject(Object node);` style, but don't feel strongly. Everything else looks great IMO

truongsinh:
Btw, all the styles are directly taken from https://github.com/flutter/flutter/blob/master/analysis_options.yaml

lib/src/cache/normalized_in_memory.dart
micimize: ah I see, [`- prefer_generic_function_type_aliases`](https://github.com/dart-lang/sdk/blob/master/docs/language/informal/generic-function-type-alias.md)

micimize:
@HofmannZ cool, I don't have merge authorization though, so you'll have to merge it

Issue #125, opened 2018-10-19
Decouple operation apis from flutter widgets


dopecoder:
**Describe the bug**
When using scoped_model in flutter, i'm unable to return subscriptions as streams as the subscriptions are implemented as widgets.

**Expected behavior**
Being able to subscribe without the use of flutter widgets.

Is there a way to just use pure dart for performing operations such as querying, subscribing and mutating as its not possible to subscribe date in scoped_model using widgets.

arnold-parge:
Check the code in my issue: [https://github.com/zino-app/graphql-flutter/issues/126](https://github.com/zino-app/graphql-flutter/issues/126)

I am subscribing without Subscription widget.

micimize:
@truongsinh [is working on this](https://github.com/zino-app/graphql-flutter/issues/196#issuecomment-476709199), and it is critical for #72.
Also from @feinstein:
> if I want to make a Query on my BLoC or Reducer or ViewModel or any form of code that's not presentation, I can't
>
> If this library can be used in a Model class, please also provide its use on the Readme. This is very helpful for unit testing and DI, where we can just simulate a query on a model class and serve it to a Presentation Widget.
 
For now I encourage anyone with urgent need to look into the widget implementations - you might be able to fit the client interaction points to whatever alternative pattern you're using. For instance, for basic `Query` and `Mutation` integration, the main things you'll have to port would be:
* `QueryOptions`, `MutationOptions`, and `WatchQueryOptions` 
* `observableQuery` management (`observableQuery = client.watchQuery(_options);`)
* collecting the `observableQuery.stream` for usage as in the `StreamBuilder`

truongsinh:
Separation MR at #203

micimize:
@truongsinh contributed a [bloc implementation example](https://github.com/zino-app/graphql-flutter/blob/next/example/lib/graphql_bloc/bloc.dart). This is similar enough to #72 that I'm going to close it in favor of there, as #203 will solve both at once

PR #202, opened 2019-03-27
 Support GraphQL Upload spec, add bloc example, tests, and coverage

### LINKED ISSUE #183: Support GraphQL Multipart Requests
We are currently unable to do things like upload images and large files directly using this library.

The library should support the multipart request spec: https://github.com/jaydenseric/graphql-multipart-request-spec

truongsinh:
Close  #183

Please pardon me because of this huge request, because I was building this feature on top of other 3 needed commits:
- 3b43a0d use stricter analysis style (#195)
- 68d2b3b and 93c14e7 refactor unit test and get test coverage (also address concerns at #125/#189)

Btw, here's the current unit test coverage: https://codecov.io/gh/truongsinh/graphql-flutter/tree/master/lib/src

Screen Shot 2019-03-27 at 11 33 00 AM


README.md
```diff
@@ -466,10 +502,14 @@ Thanks goes to these wonderful people ([emoji key](https://github.com/kentcdodds
 
 This project follows the [all-contributors](https://github.com/kentcdodds/all-contributors) specification. Contributions of any kind are welcome!
 
+[build-status-badge]: https://api.cirrus-ci.com/github/truongsinh/graphql-flutter.svg
+[build-status-link]: https://cirrus-ci.com/github/truongsinh/dart-uuid/master
+[coverage-badge]: https://codecov.io/gh/truongsinh/graphql-flutter/branch/master/graph/badge.svg
+[coverage-link]: https://codecov.io/gh/truongsinh/graphql-flutter
```
micimize: will have to remember to change these links post-merge

lib/src/core/query_manager.dart
```diff
@@ -119,8 +119,11 @@ class QueryManager {
       ).first;
 
       // save the data from fetchResult to the cache
-      if (fetchResult.data != null &&
-          options.fetchPolicy != FetchPolicy.noCache) {
+      if (
+          // should never cache a mutation
```
micimize: @truongsinh what's wrong with caching mutation results?

example/lib/main.dart
```diff
@@ -1,70 +1,39 @@
 import 'package:flutter/material.dart';
-import 'package:graphql_flutter/graphql_flutter.dart';
 
-import './mutations/mutations.dart' as mutations;
-import './queries/readRepositories.dart' as queries;
-
-const String YOUR_PERSONAL_ACCESS_TOKEN = '<YOUR_PERSONAL_ACCESS_TOKEN>';
+import 'bloc.dart' show Bloc, Repo;
```
micimize: Instead of overhauling the example to use the bloc pattern, can you add it as a separate example, like `bloc_example`?

lib/src/core/query_manager.dart
truongsinh: Similar to HTTP POST request, it should never be cached, because a mutation is a request to the server to do something, not to get information (which is the functionality of a query)

example/lib/main.dart
truongsinh: will do

lib/src/core/query_manager.dart
micimize: Hmm, I think pragmatically it's better to follow [apollo's design](https://www.apollographql.com/docs/react/advanced/caching#automatic-updates) regardless, but it also seems idiomatic according to the [graphql learn's mutation section](https://graphql.org/learn/queries/#mutations):
> Just like in queries, if the mutation field returns an object type, you can ask for nested fields. This can be useful for fetching the new state of an object after an update.

truongsinh: Thanks for pointing out those 2 documents, frankly I haven't fully understood them, but for now it's safe to reverse this change (i.e. mutation can also be cached). Later on I will have some test to ensure that even with cache, the 2 uploads mutation are still call with 2 HTTP Multipart request

micimize:
This PR:
* implements [multipart uploads]( as proposed at https://github.com/jaydenseric/graphql-multipart-request-spec)
* adds coverage and more tests
* adds a `bloc` example for alternate library usage

I think the examples could be better organized, but I want to do a larger docs/examples update later anyways

.cirrus.yml
```diff
@@ -0,0 +1,21 @@
...
+  # pub_cache:
+  #   folder: $HOME/.pub-cache
+  #   fingerprint_script: cat pubspec.lock
+  #   populate_script: pub get
+  pub_script: flutter packages get # later when we seperate graphql_client and graphql_flutter, we can change this
+  format_script: dartfmt **/*.dart -n --set-exit-if-changed
+  # analyze_script: dartanalyzer --fatal-infos .
+  publishable_script: pub publish --dry-run
+  environment:
+    CODECOV_TOKEN: ENCRYPTED[fa57a1dda8bca6ad45ef10043f632b68b7824b60c634415f741a847788c1a44ae9374155e3af3346fb9b6d72bac0e1d3]
```
micimize: @truongsinh is this sensitive?

truongsinh: This one has been encrypted with my repo, so it's not sensitive, but not working when merged to upstream. By not working, I mean the code coverage data is processed.

PR #199, opened 2019-03-26
Optimistic Updates, Lazy Dereferences, Refetch queries

micimize:
closes #127
closes #190 

This is a pretty big change. I threw in some other smaller stuff (`refetch`, `rebroadcastQueries`), but the main thing is that this implements:
* an `OptimisticCache`, which maintains a stack of patches for tracking optimistic state
  * patches can be heirachical, and this is taken advantage of by `Mutation#update`:
    > The optimistic cache layer id `update` will write to
    > is a "child patch" of the default optimistic patch
    > created by the query manager
* a `LazyCacheMap` that handles lazy dereferencing from the cache, as well as track the optimistic state of entities in the cache

### Breaking changes

- Broke `onCompleted` signature because it didn't match apollo's and is only called when `data` is ready.
- Moved `_inMemoryCache` to `@protected data` for testing/override purposes (important for `OptimisticPatches`
- Updated the example to use optimism
- adds a `refetch` argument to the `Query` `builder`

#### Fixes / Enhancements

- Added `OptimisticCache` and related attributes to `QueryResult` (`optimistic`, `timestamp`) 
- Added `lazy_cache_map.dart` for handling cyclical dereferences in the normalized cache
  - added `CacheState` for tracking optimism from the perspective of normalized cache entities
- Added `raw_operation_data.dart` to consolidate base functionality
- Added `rebroadcastQueries` to the `QueryManager`, for use post-update, which rebroadcasts all "safe" queries that can be with updated data from the cache
- Added `optimisticResult` management to the `QueryManager`
- Added `optimisticResult` to `BaseOptions`, and `QueryOptoins` (it is added in `runMutation` for mutations)
- Added `optimistic` attribute `QueryResult` itself for lifecycle management. 

#### Docs
* `LazyCacheMap` usage and reasoning
*  Optimism section. differences between `result.optimistic` and `LazyCacheMap.isOptimistic`
* `update`, `onCompleted` usage/existence
* `refetch` usage/existence

Optimism should still be considered experimental. The biggest question requiring further investigation is what the relationship between `FetchPolicy` and optimism is in apollo.

To give a more detailed survey of how optimism works in a `mutation` update situation:
* if provided to `runMutation`, an `optimisticResult` is added to `QueryOptions`
* this is handled by the `QueryManager`, which then:
  * writes the optimistic data as a patch with the id `$queryId` to the `OptimisticCache`
  * adds a wrapping `optimistic` `QueryResult` to the query stream
[...]

PR #203, opened 2019-03-27
Fix #72/#125/#189, separate into 2 package while remain monorpoe

### LINKED ISSUE #72: Separate generic GraphQL client code from flutter implementation
When trying to use Graphql-flutter in a framework-agnostic way, it is today possible to target directly non-Flutter specific parts (Client, ...) of this library, which is probably the best in Dart as of today.

For example, following the BLoC pattern, we use graphql-flutter inside our business logic components that (even if used today with Flutter) are intended to be framework-agnostic (and could be used with AngularDart for example ?).

The problem is that, making a separate Dart package for this business logic components, a dependency on graphql-flutter would bring a dependency on Flutter itself, even if we're not using Flutter-specific parts of the package.

Separate this wonderfull lib in 2 packages : 1 framework-agnostic with generic GraphQL, and another one with Flutter implementation targetting the first one

If this is not intended, we would probably create a new library based on the "generic" GraphQL implementation from this library. This would probably bring some overhead in maintaining synchronization between the 2 projects contributions.

### LINKED ISSUE #125: Decouple operation apis from flutter widgets

truongsinh:
Fix #72/#125/#189

Separate into 2 packages, `graphql_client` (I was added as the "uploader" for this dart pub package) and `graphql_flutter`.


micimize:
This PR is pretty massive, might take a bit for me to get to it. Will be easier to look at once #202 is merged. Also, merging it would make merging downstream work pretty difficult. Might be best to defer merging until we've consolidated more features like optimism (#199) and do a big rebase of this PR.

micimize:
Apologies, but I'm closing this in favor #245. I'm sure this PR was done just fine, it's just that I wanted to merge the prior outstanding PRs before it, and after that was done I didn't want to try and rebase these diffs 😅

PR #210, opened 2019-04-03
Check operationName before constructing regex in Operation.isSubscription, fixes #186

### LINKED ISSUE #186: ": Undefined location" when running a manual Mutation (next branch) + Random 127.0.0.1 connections
**Describe the bug**
When running a mutation manually, I get 2 errors at different code levels:
- `SocketException: OS Error: Connection refused, errno = 111, address = localhost, port = 33860` with random port (location is: `flutter_graphql/src/link/link_http.dart:85`) ... this is directly related to the mutation that I'm trying to run
- `: Undefined location` after getting the awaited result of the mutation. Note that my backend is on localhost.

I was trying to get to the code that caused connections to random ports, but I couldn't use the IDE to pin-point the source of the problem. Makes me wonder, why is it doing this random connection and not connecting to where it's supposed to.

**To Reproduce**
1. Start any project that would use `flutter_graphql` (same exact behavior when using `graphql_flutter`'s `next` branch or 1.0.0 beta)
2. Use an `HttpLink` that is pointing to a fixed `http://localhost:8080/v1/graphql`
3. At this point, any manual mutations I tried to run would cause the same exact error
4. Even absolutely invalid document will also return the same error
5. Set a break-point deeper and deeper, and eventually, you'll meet `SocketException`

**Expected behavior**
A connection is made to my GraphQL backend. Unfortunately, my backend stays still, not receiving any connection at all.

**Screenshots**
Here are some screens:

image

image

 - OS: Linux Mint (based on Ubuntu)
 - Version: 19.1

**Smartphone (please complete the following information):**
 - Device: Emulator
 - OS: Android
 - Version: any

**Additional context**
I was looking at the 0.9.5 version, but there was no way to run manual Mutations.

yunyu:
On the `next` branch, manual mutations fail because of the newly added isSubscription getter throwing an exception if operationName is null (which seems to be the case for non-subscriptions). Adding a nullcheck fixes the issue (at least for me) when using `GraphQLClient.mutate`.

Fixes https://github.com/zino-app/graphql-flutter/issues/186

### DIFF
--- a/lib/src/link/operation.dart
+++ b/lib/src/link/operation.dart
@@ -29,7 +29,9 @@ class Operation {
     return result;
   }
 
-  bool get isSubscription => document.contains(RegExp(r'.*?subscription ' + operationName));
+  bool get isSubscription =>
+      operationName != null &&
+      document.contains(RegExp(r'.*?subscription ' + operationName));
 
   String toKey() {
     /// SplayTreeMap is always sorted


micimize:
Great! Hopefully this is the main root cause.

Pushing these through, but @yunyu if you have the chance I'd recommend setting up [commit signing](https://help.github.com/en/articles/about-commit-signature-verification)

PR #213, opened 2019-04-09
disable polling when interval is set to zero

mainawycliffe:
Disables polling interval when set to zero, this caused it to poll the GraphQL server nonstop. 

#### Fixes / Enhancements

- Added a check at startPollingQuery method to check whether polling is zero, if zero, it doesn't register it for polling


mainawycliffe:
Yeah, that makes more sense. Let me make the changes and get back to you.

micimize:
@mainawycliffe pushing this through, but if you get the chance set up signature verification - https://help.github.com/en/articles/about-commit-signature-verification

mainawycliffe:
Thanks. I will.

Issue #238, opened 2019-04-24
Compile error on Flutter 1.5.5


mateusfsilva:
**Describe the bug**
After update flutter to version 1.5.5 the following error is happening on run:

```
Compiler message:
file:///Users/mateus.gustavo/desenvolvimento/flutter/.pub-cache/hosted/pub.dartlang.org/graphql_flutter-1.0.0+3/lib/src/widgets/subscription.dart:15:3: Error: Function-typed
parameters can't specify 'const', 'final' or 'var' in place of a return type.
Try replacing the keyword with a return type.
  final bool loading,
  ^^^^^
file:///Users/mateus.gustavo/desenvolvimento/flutter/.pub-cache/hosted/pub.dartlang.org/graphql_flutter-1.0.0+3/lib/src/widgets/subscription.dart:16:3: Error: Function-typed
parameters can't specify 'const', 'final' or 'var' in place of a return type.
Try replacing the keyword with a return type.
  final T payload,
  ^^^^^
file:///Users/mateus.gustavo/desenvolvimento/flutter/.pub-cache/hosted/pub.dartlang.org/graphql_flutter-1.0.0+3/lib/src/widgets/subscription.dart:17:3: Error: Function-typed
parameters can't specify 'const', 'final' or 'var' in place of a return type.
Try replacing the keyword with a return type.
  final dynamic error,
  ^^^^^
```

**To Reproduce**
1. `flutter channel dev`
2. `flutter upgrade`
3. `flutter run` or `flutter test`

**Expected behavior**
The project needs to run without compile errors.

**Desktop:**
 - OS: macOS High Sierra
 - Version 10.13.6

**Additional context**
To fix the issue just delete the `final` keyword from the parameters of the `SubscriptionBuilder` typedef.

```dart
typedef SubscriptionBuilder<T> = Widget Function({
  bool loading,
  T payload,
  dynamic error,
});
```

psyanite:
I have the same issue

mainawycliffe:
@psyanite  Are you too using dart 2.3?

psyanite:
Flutter 1.5.5 and 1.5.4 failed. 1.5.3 worked.

mainawycliffe:
It seems this a bug with dart 2.3. @mateusfsilva has a PR on the same, you can try it out.

micimize:
merged and should be fixed in `1.0.0+4` (tested on my own depending app)

[...]

PR #239, opened 2019-04-24
Fix issue #238. Compile error on Flutter 1.5.5

mateusfsilva:
Fix issue #238.

#### Fixes / Enhancements

- Fixed the compile error on Flutter 1.5.5. The error was the `final` keyword on `SubscriptionBuilder` typedef.


micimize:
Even if this is somehow a regression, this signature isn't correct anyhow.

@mateusfsilva thanks for fixing your own issue!

I recommend setting up [commit signing](https://help.github.com/en/articles/about-commit-signature-verification) going forward

Issue #189, opened 2019-03-17
Provide methods for Dart and not only Flutter. 


feinstein:
I didn't use this library, I am just starting to look for GraphQL in Flutter, but looking at the Readme, this library seems to be really tight with Flutter Widgets, so if I want to make a Query on my BLoC or Reducer or ViewModel or any form of code that's not presentation, I can't.

If this library can be used in a Model class, please also provide its use on the Readme. This is very helpful for unit testing and DI, where we can just simulate a query on a model class and serve it to a Presentation Widget.

micimize:
Consolidating this with #125

PR #245, opened 2019-04-27
Split client into stand-alone and flutter packages

### LINKED ISSUE #72: Separate generic GraphQL client code from flutter implementation

micimize:
Closes #72, #125, #189

Splits out the core `dart` functionality into `packages/graphql` for use as a stand-alone client 
```dart
import 'package:graphql/client.dart';
// for ObservableQuery and other things used by graphql_flutter
import 'package:graphql/internal.dart';
```

`graphql_flutter` should not be broken in any way

### Breaking changes

- Broke directory structure to support multi-package development

#### Fixes / Enhancements

- Added `packages/graphql`

#### Docs

Severely lacking


micimize:
The docs for the standalone `graphql/client.dart` are still severely lacking, but I'm currently quite busy with other projects. I would prefer to get this merged, and publish something like `graphql 1.0.0-alpha-dev`.

Issue #237, opened 2019-04-23
Implement 'Keep a Changelog'


HofmannZ:
When releasing a new version it's always hard to deliver a proper changelog. [Keep a changelog](https://keepachangelog.com/en/1.0.0/) is a proposal on how we can deliver a consistent and complete changelog.

I propose we implement it in this repo. I would require the following steps:
- [ ] Refactor the changelog to the [keep a changelog spec](https://keepachangelog.com/en/1.0.0/).
- [ ] Add update the [pull request template](https://github.com/zino-app/graphql-flutter/blob/master/.github/PULL_REQUEST_TEMPLATE.md) to mention the changelog requirements.

mainawycliffe:
Is it possible to automatically parse changes from PRs, so we can have a section for change log in the PR?

HofmannZ:
For PRs; I’m not sure, but for [branches it does work](https://github.com/semantic-release/semantic-release). It does requires better commit messages so that’s a possible bottle neck.


The flow would be:
- Merge to beta branch: create a beta Github version tag.
- Merge to master: create a stable Github version tag.

Then we can add a CI trigger on git tags to do a `pub publish`.

The only problem is that it can’t deal with the `+1` in the Dart/Flutter version tags.

So it might not be the best fit..

micimize:
closed long ago

Issue #91, opened 2018-09-06
Include tests and implement CI


HofmannZ:
**Is your feature request related to a problem? Please describe.**
There currently is no way to programatically determine if pushed commits still provide a stable API. We require a peer for each PR, but some bugs might still slip trough.

**Describe the solution you'd like**
I suggest to create Unit Tests with **a least 80%** coverage. In addition I suggest to implement an automated build and deploy pipeline.  

**Additional context**
- [x] Write tests
- [x] Setup automated builds

PR #247, opened 2019-04-28
chore: implement ci and cd

HofmannZ:
Solves #237 and #91.

In this PR introduces a CI/CD setup with automated releases. It does require us to uses [Angular style](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#-git-commit-guidelines) commit messages. So we should maybe add some contributor guidelines.

#### Dependencies:

Installs the dependencies and saves it to the cache for the following steps.

#### Lint:

Lints the source code using `dartfmt`.

#### Coverage:

Runs the tests and generates test coverage reports.

#### Release:

The [semantic-release](https://github.com/semantic-release/semantic-release) package automatically read the commit messages since the last release and determines the version number based on it. And generates a Github release from it.

#### Publish:

Patches the version int the `pubspec.yaml` and publishes the package to Pub.


pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: graphql_flutter
 description: A GraphQL client for Flutter, bringing all the features from a modern GraphQL client to one easy to use package.
-version: 1.0.0+4
+version: 0.0.0-development
```
eusdima: Why are we starting from 0.0.0?

HofmannZ: The version gets overwritten, by the deploy job. So to avoid having to commit the version via the CI, I removed it altogether and let the semver be leading.

HofmannZ:
Do we want to deploy a beta version alongside the stable version? If so, I can do that automatically.

pubspec.yaml
micimize: I think keeping the version number in sync would be a nice to have, so I made #250 to track it separately

packages/graphql_flutter/pubspec.yaml
```diff
@@ -25,4 +25,4 @@ dev_dependencies:
   test: ^1.5.3
 
 environment:
-  sdk: ">=2.2.0 <3.0.0"
+  sdk: ">=2.1.0 <3.0.0"
```
micimize: Proof that CI is working
> Because graphql_flutter depends on graphql >=1.0.1-beta which requires SDK version >=2.2.0 <3.0.0, version solving failed.

HofmannZ:
@eusdima Done!

@micimize I agree, it's changed.

HofmannZ:
@micimize I will be waiting for your review before I merge.

PR #248, opened 2019-04-28
Implement error policy

mainawycliffe:
Attempt to implement errorPolicy, to address this [issue](https://github.com/zino-app/graphql-flutter/issues/143)

### Breaking changes

- No breaking changes, should work expected without code being updated!

### DIFF
--- a/packages/graphql/lib/src/core/query_manager.dart
+++ b/packages/graphql/lib/src/core/query_manager.dart
@@ -280,7 +280,7 @@ class QueryManager {
 
     // check if there are errors and implement error policy
     // add errors if the [ErrorPolicy] allows to report errors i.e. it errorPolicy != ignore
-    if (_shouldReturnError(fetchResult, options.errorPolicy)) {
+    if (fetchResult.errors != null && _shouldReturnError(options.errorPolicy)) {
       errors = List<GraphQLError>.from(fetchResult.errors.map<GraphQLError>(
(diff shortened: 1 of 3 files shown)


packages/graphql/lib/src/core/query_manager.dart
```diff
@@ -269,23 +271,61 @@ class QueryManager {
...
+
+  bool _shouldReturnData(FetchResult fetchResult, ErrorPolicy errorPolicy) {
+    // if there are no errors, return data
+    if (fetchResult.errors == null) {
+      return true;
+    }
+
+    // if there is data and errors, return data if the error policy allows it
+    // if errorPolicy is return all or ignore errors, return data
+    if (fetchResult.errors != null &&
```
micimize: for this and in `_shouldReturnErrors` I think a switch statement would be clearer:
```dart
switch (options.errorPolicy) {
  case ErrorPolicy.ignore:
  case ErrorPolicy.all:
    return false;
  case ErrorPolicy.none:
  default:
    return true;
}
```
but that's a pretty minor thing - overall this is great and I'm green lighting it 👍

micimize:
@mainawycliffe leaving open for a bit if you want to make the `switch`, otherwise just comment and I'll merge

mainawycliffe:
@micimize yes, i agree switch case is much clearer. I will make the necessary changes.

PR #255, opened 2019-04-30
chore: include a licence in the root of the project

HofmannZ:
There were two licence files both not in the root of the project. This PR suggests moving the licence file to the root of the project.


micimize:
I think we should actually have the LICENSE copied to top and package levels, like https://github.com/dart-lang/json_serializable and https://github.com/felangel/bloc

HofmannZ:
@micimize I agree and changed it.

PR #267, opened 2019-05-10
feat(graphql) change Link

klavs:
(Actually, just playing around with Dart)

Some changes/improvements to `Link`.
Making `request` a method on `Link` appears to improve code layout.

### Breaking changes

Links can now be implemented either by extending `List` or using `RequestHandler` via `Link.fromHandler`.

#### Fixes / Enhancements

- make `request` a method on `Link`
- refactor existing links accordingly
- implement `Link.from` to join links in sequence
- implement `Link.fromHandler` to create a `Link` using `RequestHandler`

### DIFF
--- a/packages/graphql/lib/src/link/link.dart
+++ b/packages/graphql/lib/src/link/link.dart
@@ -12,26 +12,42 @@ typedef RequestHandler = Stream<FetchResult> Function(
   NextLink forward,
 ]);
 
-Link _concat(
-  Link first,
-  Link second,
-) {
-  return Link(request: (
-    Operation operation, [
-    NextLink forward,
-  ]) {
-    return first.request(operation, (Operation op) {
-      return second.request(op, forward);
-    });
-  });
+abstract class Link {
+  static Link from(List<Link> links) => links.reduce(
+        (first, next) => first.concat(next),
+      );
+
+  static Link fromHandler(RequestHandler handler) =>
+      _RequestHandlerLink(handler);
+
+  Stream<FetchResult> request(Operation operation, [NextLink forward]);
+
+  Link concat(Link next) => _ConcatLink(this, next);
 }
 
-class Link {
-  Link({this.request});
+class _RequestHandlerLink extends Link {
+  _RequestHandlerLink(this._handler);
 
-  RequestHandler request;
+  RequestHandler _handler;
 
-  Link concat(Link next) => _concat(this, next);
+  @override
+  Stream<FetchResult> request(Operation operation, [NextLink forward]) {
+    return _handler(operation, forward);
+  }
+}
+
+class _ConcatLink extends Link {
+  _ConcatLink(this._first, this._second);
+
+  Link _first;
+  Link _second;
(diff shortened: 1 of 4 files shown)


HofmannZ:
Could you elaborate on why this is better? Since it's a breaking change I'd like some more explanation why we need it.

klavs:
Oh, I was under impression that the `beta` branch is a breaking change...
I guess you can close this PR if that is not the case.

My rationale is as follows:

I cannot argue that this is provably better. I can only say that
- the current implementation is not consistent with Apollo, where you can either extend `Link` overriding it's `request` method and pass request handler to the constructor
- in the current implementation, if you want to extend a `Link` (which is how all three links in the repo are implemented) you are forced to either inline huge `request` function or map `request` function to a method
- if we cannot perfectly match Apollo, we can choose to implement `Link` in a way that simplifies both approaches (extending, request handler)
- using extension it is natural to expect that you can directly override a method (`request`)
- ability to use function as a link can be achieved via static function `List.fromHandler`

packages/graphql/lib/src/link/link.dart
```diff
@@ -12,26 +12,42 @@ typedef RequestHandler = Stream<FetchResult> Function(
...
+
+  static Link fromHandler(RequestHandler handler) =>
+      _RequestHandlerLink(handler);
+
+  Stream<FetchResult> request(Operation operation, [NextLink forward]);
+
+  Link concat(Link next) => _ConcatLink(this, next);
+}
+
+class _RequestHandlerLink extends Link {
```
micimize: The spirit of this is good, but we can do this in both a completely backwards-compatible and more Apollo compliant way:
* `_RequestHandlerLink` was the publicly exposed `Link`, with an api like
```dart
Link({RequestHandler request}): this._handler = request
```
* dump all the `abstract class Link` static methods into `_RequestHandlerLink`
* add some safety in `request` that's something like
```dart
assert(_handler != null, "Custom Link $runtimeType has no RequestHandler")
```

packages/graphql/lib/src/link/link.dart
```diff
@@ -12,26 +12,42 @@ typedef RequestHandler = Stream<FetchResult> Function(
...
+
+  @override
+  Stream<FetchResult> request(Operation operation, [NextLink forward]) {
+    return _handler(operation, forward);
+  }
 }
 
-class Link {
-  Link({this.request});
+class _ConcatLink extends Link {
```
micimize: This seems like overkill - let's stick to `_concat`. What `concat` really needs is [more safety / warnings](https://github.com/apollographql/apollo-link/blob/089f71be5ea6a2ccd465751a500b16b97dbc23aa/packages/apollo-link/src/link.ts#L62)

HofmannZ:
@klavs We try to keep the beta branch not breaking when it does not have to be. The rationale behind your PR is good, and in the spirit of not breaking things, I believe that @micimize suggestions are a great way to get the changes into the library.

Issue #264, opened 2019-05-07
Certain queries fail due to parser issue


Fhffsthu:
There is an issue in the graphql_parser with certain queries - see [https://github.com/angel-dart/graphql/issues/23](https://github.com/angel-dart/graphql/issues/23) for the issue and sample queries. 

The parsing failure causes the "operationName" to be set to null. In the 1.0.0+4 (as opposed to 1.0.0-beta.1+1) this has been detected and fixed by changing a null "operationName" to `'UNNAMED/' + document.hashCode.toString();`. This seems to guarantee the query will fail. Commented out this line seems to allow the queries to work - though I am guessing it will break something else.
 
This query will fail:
```
query customerMemberAttributes($memberId: Int!){
  customerByCustomerId(customerId: 7) {
    customerMemberAttributesByCustomerId {
      nodes {
        customerMemberAttributeId
        memberAttributesByCustomerMemberAttributeId(condition: {memberId: $memberId}) {
          nodes {
            memberAttributeId
          }
        }
      }
    }
  }
}
```
this query does not fail:
```
query customerMemberAttributes($memberId: Int!){
  customerByCustomerId(customerId: 7) {
    customerMemberAttributesByCustomerId {
      nodes {
        customerMemberAttributeId
        memberAttributesByCustomerMemberAttributeId(condition: {memberId: 7}) {
          nodes {
            memberAttributeId
          }
        }
      }
    }
  }
}
```

I am unsure what issue the check for the null query name fixed but the current version of the library does fail a few of my queries.

endigo:
@HofmannZ Related to #235 

I have the same problem.

micimize:
(hopeflly) closed by #276

PR #276, opened 2019-05-19
Handle Anonymous ops

### LINKED ISSUE #235: [Beta 2] query keyword is required in all queries
**Describe the bug**
Previously, you could make a query like so:
```dart
const String getUsers = """
  {
    users {
      id
      name
    }
  }
"""
```

however, now the `query` keyword is required, even if no variables are being passed.

```dart
const String getUsers = """
  query getUsers {
    users {
      id
      name
    }
  }
"""
```

**To Reproduce**
1. Attempt to make the first query (any query without `query`
2. *Failure*

**Expected behavior**
`query` should not be required.

micimize:
closes #235, #264

### Breaking changes

- made `operationName` nullable because anonymous operations are allowed 

#### Fixes / Enhancements

- Fixed anonymous `{ foo  { bar } }` queries (were broken)
- Added an `_identifier` to `RawOperationData` (maybe a bit redundant with the document in the `toKey` as well
- Cleaned up unit tests some

### DIFF
--- a/packages/graphql/lib/src/core/raw_operation_data.dart
+++ b/packages/graphql/lib/src/core/raw_operation_data.dart
@@ -24,14 +24,20 @@ class RawOperationData {
   Map<String, dynamic> variables;
 
   String _operationName;
+  String _documentIdentifier;
 
   /// The last operation name appearing in the contained document.
   String get operationName {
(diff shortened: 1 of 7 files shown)


mainawycliffe:
LGTM

PR #286, opened 2019-05-20
Starwars example

micimize:
* Adds a locally-runnable example app to `examples/`
* Fix `Query.refetch`

#### Fixes / Enhancements
- Moved `refetchQuery` to manager because scheduler  doesn't see non-subscriptions
- Add a full client/server example `examples/starwars` with subscriptions


mainawycliffe:
This Looks and works great. The only change i had to make was for the server ip address, from `127.0.0.1` to `10.0.2.2` when using Android Emulator on windows. Don't know how this works on other OSs but may be  add a note on the read me file.

micimize:
@mainawycliffe good catch - added a `Platform` toggle

PR #293, opened 2019-05-23
docs: add variable type to query example

pleopardi:
#### Docs

- Added missing variable type in query example


mainawycliffe:
Hi, thanks for the PR. Can you please make your PR against beta branch instead of master. And also the commit message is should be docs not refactor, just like your PR title, because you are making a change against the docs.

pleopardi:
Thanks for the suggestions @mainawycliffe, now it should be fine!

PR #302, opened 2019-05-28
first widget test for `graphql_flutter` package

truongsinh:
very first widget test for `graphql_flutter` package. When doing so, I happen to discover that we are not null-safe in some places

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,15 +101,16 @@ jobs:
             pub global activate coverage
             pub get
             pub run test_coverage
-      # - run:
-      #     name: Code coverage (graphql_flutter)
-      #     command: |
-      #       cd packages/graphql_flutter
-      #       flutter test --coverage
+      - run:
+          name: Code coverage (graphql_flutter)
+          command: |
+            cd packages/graphql_flutter
+            flutter test --coverage
       - run:
           name: Upload coverage
           command: |
-            bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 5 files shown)


packages/graphql/pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: graphql
 description: A stand-alone GraphQL client for Dart, bringing all the features from
   a modern GraphQL client to one easy to use package.
-version: 1.0.2-beta.2
+version: 1.1.0
```
HofmannZ: Can you please change this back?

truongsinh: ok, sorry i did't pay attention to this.

truongsinh: ah no, I meant to merge this to `master`, not `beta`. It seems that I can't change the destination, so I will close this and make another PR.

mainawycliffe:
We should be making this changes against beta, not master. But looks good. https://github.com/zino-app/graphql-flutter/blob/master/CONTRIBUTING.md

truongsinh:
> We should be making this changes against beta, not master. But looks good. https://github.com/zino-app/graphql-flutter/blob/master/CONTRIBUTING.md

Sorry I took a long gap 😅. Anyway, I don't understand why we should target `beta` branch (I'm a fan of https://trunkbaseddevelopment.com/), but let's take that discussion in another thread. I will make neccessary change.

mainawycliffe:
LGTM

Issue #221, opened 2019-04-18
Switch to pedantic style 


micimize:
> general purpose and flutter specific libraries should use `package:pedantic`  

@truongsinh we have [clarification on `pedantic` usage](https://github.com/dart-lang/pedantic/issues/17#issuecomment-484398115). I think it'd be best to follow the ecosystem on this one

micimize:
closed by #303

PR #303, opened 2019-05-28
use `pedantic` style `flutter analyze` for CI

truongsinh:
No breaking change, implement #221 with CI enforcement


packages/graphql_flutter/pubspec.lock
```diff
@@ -91,10 +91,10 @@ packages:
   graphql:
     dependency: "direct main"
     description:
-      name: graphql
-      url: "https://pub.dartlang.org"
-    source: hosted
-    version: "1.0.1-beta.8"
+      path: "../graphql"
```
micimize: Do we need to regen this lock file so it doesn't have a relative dependency?

packages/graphql/pubspec.yaml
```diff
@@ -18,7 +18,7 @@ dependencies:
   graphql_parser: ^1.1.3
   rxdart: ^0.22.0
 dev_dependencies:
-  perfectionist: ^2.0.0
+  pedantic: "<=1.7.99"
```
HofmannZ: Why not latest?

packages/graphql_flutter/pubspec.lock
mainawycliffe: May be we should just add the pubspec.lock files to gitignore https://dart.dev/guides/libraries/private-files#pubspeclock

packages/graphql/pubspec.yaml
truongsinh: the latest is conflict with `flutter_test` package, which requires exactly `1.5.0` (i tried and the dependency resolver failed)

packages/graphql_flutter/pubspec.lock
truongsinh: will ignore as suggested

PR #304, opened 2019-05-29
Fix LegacyInitOperation `toJson`

truongsinh:
`toJson()` should always return a `<String, dynamic>` not something else, especially not `String` (`InitOperation` is working fine, only `LegacyInitOperation` is broken)

Fix https://github.com/zino-app/graphql-flutter/issues/134

After merging to master, we should pump patch version so that people can use this fix.

### DIFF
--- a/packages/graphql/lib/src/websocket/messages.dart
+++ b/packages/graphql/lib/src/websocket/messages.dart
@@ -33,7 +33,7 @@ class MessageTypes {
 }
 
 abstract class JsonSerializable {
-  dynamic toJson();
+  Map<String, dynamic> toJson();
 
   @override
   String toString() => toJson().toString();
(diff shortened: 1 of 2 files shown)


packages/graphql/lib/src/websocket/messages.dart
```diff
@@ -74,15 +74,15 @@ class LegacyInitOperation extends InitOperation {
...
+  Map<String, dynamic> toJson() {
     final Map<String, dynamic> jsonMap = <String, dynamic>{};
     jsonMap['type'] = type;
 
     if (payload != null) {
       jsonMap['payload'] = json.encode(payload);
     }
 
-    return json.encode(jsonMap);
+    return jsonMap;
```
mainawycliffe: Why are we removing JSON encoder?

truongsinh: as explain the description of this PR, toJson() should always return a <String, dynamic> not something else, especially not String.

PR #311, opened 2019-05-31
Chore/skip browser test

truongsinh:
skip 3 tests on browser platform, relating to the effort to make this package compatible to web/browser #295.

You can test this PR by running `pub run test test -p chrome` and `pub run test test -p vm`.

`pub run test test -p chrome` is not yet supported on CI, see https://github.com/cirruslabs/docker-images-flutter/issues/17

Issue #295, opened 2019-05-24
make `graphql/client.dart` web compatible


micimize:
We've received a handful of complaints about the `dart:io` dependency, which makes `graphql/client.dart` [incompatible](https://api.dartlang.org/stable/2.3.1/dart-io/dart-io-library.html) with web:

> **Important**: Browser-based applications can't use this library. Only servers, command-line scripts, and Flutter mobile apps can import and use dart:io.

This should be avoidable.

Current progress and corresponding MR:

- [x] memory persistent #325
- [x] GraphQL upload file #328
- [x] websocket #323

micimize:
`dart:io` is used:
* in `lib/src/cache/in_memory.dart` for persistence
* in `lib/src/core/raw_operation_data.dart` for `File` handling
* in `lib/src/link/http/link_http.dart` for `File` upload logic
* in `lib/src/socket_client.dart` for `CompressionOptions`, `WebSocket`, and `WebSocketStatus`

micimize:
We can probably replace file logic with https://pub.dev/packages/file, but it seems like `WebSockets` will require some toggle logic between `dart:io` and `dart:html`. I think we can do a [conditional export](https://github.com/sass/dart-sass/blob/3b6730369bdff9bb7859411c82e4b21f194d2514/lib/src/io.dart)

adelcasse:
Considering Websocket (that's our pain point with @kei42 for now), I've come across this lib https://github.com/dart-lang/web_socket_channel

Maybe a solution would be to use it or take inspiration to wrap both dart:io and dart:html Websocket implementations ?

micimize:
@adelcasse that's the right direction - maybe something more minimalist, idk much about websockets though. I'm not sure exactly, but I think we can leverage the conditional export to avoid any api differences between the platform (as opposed to shared interface with injection).

It'll probably be a bit before I have the time to get to this

adelcasse:
beware on conditional import as I think the implementations on io and html are not the same (different methods, ...). @kei42 might confirm as he tried but we have pull different code conditionally, not only imports

kei42:
Indeed as far as I've seen there is no "connect" method in dart:html and there are no stream handling either. 
 To connect on dart:html you just need this line:
`WebSocket(url,
          protocols: protocols);`
where as for dart:io 
`WebSocket.connect(url,
          protocols: protocols,
          headers: headers,
          compression: config.compression);`

micimize:
[...]

PR #312, opened 2019-05-31
avoid mutate input data

truongsinh:
Should be merged only after #311.

increase test coverage on `dart:io` related, relating to the effort to make this package compatible to web/browser #295.

While doing this, I happen to discover a bug that we are currently mutating input data

### DIFF
--- a/packages/graphql/lib/src/utilities/helpers.dart
+++ b/packages/graphql/lib/src/utilities/helpers.dart
@@ -36,14 +36,14 @@ Map<String, dynamic> _recursivelyAddAll(
   Map<String, dynamic> target,
   Map<String, dynamic> source,
 ) {
-  target = unwrapMap(target);
+  target = Map.from(unwrapMap(target));
   source = unwrapMap(source);
   source.forEach((String key, dynamic value) {
     if (target.containsKey(key) &&
         target[key] is Map &&
         value != null &&
         value is Map<String, dynamic>) {
-      _recursivelyAddAll(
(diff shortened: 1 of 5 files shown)


mainawycliffe:
@truongsinh this PR has a commit from another PR, on [tests](https://github.com/zino-app/graphql-flutter/pull/312/commits/8e9aef8d184cc855820646f75e19043f51991f36)

truongsinh:
@mainawycliffe it's true that it depends on #311. My plan is to wait for #311 to be merged, and then merge this one, and then merge #313. I try to break my work as small as possible but still meaning full for easier review.

mainawycliffe:
if they are related, i think you should consolidate them. It's much easier plus the approval process and merge process is quite different. I can review and approve your changes but I can't merge them, so essentially i will end up reviewing the same commit multiple times.

HofmannZ:
@truongsinh I’m really a fan of small PRs 👍🏼

packages/graphql/lib/src/utilities/helpers.dart
```diff
@@ -36,14 +36,14 @@ Map<String, dynamic> _recursivelyAddAll(
   Map<String, dynamic> target,
   Map<String, dynamic> source,
 ) {
-  target = unwrapMap(target);
+  target = Map.from(unwrapMap(target));
```
micimize: good catch - I though prepending an empty literal in `deeplyMergeLeft` would do this, but it doesn't because of the deep traversal.

PR #313, opened 2019-05-31
Chore/file system exception

truongsinh:
Should be merged only after #312.

increase test coverage on `dart:io` related, relating to the effort to make this package compatible to web/browser #295. This commit is about `FileSystemException`.

While doing this, I happen to discover that in `in_memory` code and some places in our test are using `sync` variation of file/dir operations, all of which have been changes to async.


truongsinh:
pinging @micimize

PR #314, opened 2019-05-31
test(client): remove `mirror`

truongsinh:
dart test, either in flutter or web browser, does not support mirror,
see https://github.com/dart-lang/sdk/issues/30538 and
https://github.com/flutter/flutter/issues/1150.
Using `Directory.current.path` might still be problematic for web browser later on,
but now let's focus on flutter only.


micimize:
@HofmannZ lack of `pubspec.lock` seems to be breaking CI

truongsinh:
closing in favor of #323 .

PR #317, opened 2019-05-31
fix(example/starwars):zino-app#308 using toString()

### LINKED ISSUE #308: uuidFromObject would throw a error in example
**Describe the bug**
when:
object['id'] is int,it will throw a error
where it is: 
```
graphql-flutter/examples/starwars/lib/client_provider.dart 
final String id = object['id'] as String;  // line 7
```
**Expected behavior**
final String id = "${object['id']}";  // line 7

huang12zheng:
#### Fixes / Enhancements

- Fixed #308


mainawycliffe:
@truongsinh I agree with you especially on packages but am just wondering whether unit tests that necessary for examples?

huang12zheng:
> I appreciate the change and the attempt to make the PR, but we should make the habbit of fixing must be accompanied unit test.
> 
> That being said, I understand it might be overwhelming for new comers, thus I might continue on this branch and help with the unit test.

May be a workflow vedio is useful for new comers.
And unittest is very useful,So I agree with you.

huang12zheng:
>  whether unit tests that necessary for examples?

If example run unittest of packages[graphql_flutter], #308 would never happen. ^_^

huang12zheng:
I would also suggest that examples separate from here(graphql_flutter)

micimize:
I think ideally we can enforce package-level bug fixes including a test for the bug they're fixing, but only once we've made that process easy.

Like, if someone discovers and fixes a bug in subscriptions, to add a unit test they'd have to go and make all the testing fixtures for that swath of code as well.

For now this is super minor and I'ma just push it through.

PR #323, opened 2019-06-01
feat(client): websocket available on web

truongsinh:
This is an ongoing effort to make this package compatible to web/browser #295.

### Breaking changes

- SocketClient no longer accepts "headers", see https://stackoverflow.com/a/4361358/1326678. It is done so that we can support both version of WebSocket on `dart:io` and `dart:html` (web browser)

#### Fixes / Enhancements

This enhancement comes with an integration test marked with `tags: "integration"`, so if anyone want to run quick unit test only, do so with
- `pub run test -x integration`
- to run integration only,  `pub run test -t integration`
- to run integration only, on vm,  `pub run test -t integration -p vm`
- to run integration only, on browser (in this case Chrome),  `pub run test -t integration -p chrome`


packages/graphql/lib/src/link/web_socket/link_web_socket.dart
```diff
@@ -70,7 +59,7 @@ class WebSocketLink extends Link {
   void connectOrReconnect({Map<String, dynamic> headers}) {
     _socketClient?.dispose();
     _socketClient =
```
mainawycliffe: I think this is what is causing the linting test to fail

mainawycliffe: There are some more in the `socket_client.dart`

mainawycliffe:
Looks Good, just fix at the linting issues.

micimize:
I don't think we should hobble the `dart:io` websocket api just because the browser api is broken. We should maybe just throw an error in a `dart:html` websocket wrapper if there are custom headers and say it's unimplemented/unsupported.

If we look at the [hasura docs](https://docs.hasura.io/1.0/graphql/manual/subscriptions/index.html#communication-protocol), they talk about [apollo's `connectionParams`](https://github.com/apollographql/subscriptions-transport-ws/blob/cf3f492f93aa1df812f630327dc5eea33999fb71/src/client.ts#L559) (same as our `initPayload`) and then the cookies approach, which a `dart:io` client wouldn't be able to replicate without manaully setting headers, I don't think.

truongsinh:
> I don't think we should hobble the `dart:io` websocket api just because the browser api is broken. We should maybe just throw an error in a `dart:html` websocket wrapper if there are custom headers and say it's unimplemented/unsupported.
> 
> If we look at the [hasura docs](https://docs.hasura.io/1.0/graphql/manual/subscriptions/index.html#communication-protocol), they talk about [apollo's `connectionParams`](https://github.com/apollographql/subscriptions-transport-ws/blob/cf3f492f93aa1df812f630327dc5eea33999fb71/src/client.ts#L559) (same as our `initPayload`) and then the cookies approach, which a `dart:io` client wouldn't be able to replicate without manaully setting headers, I don't think.

Considering `WebSocket` is from `Web`, I don't think we can judge the spec (https://tools.ietf.org/html/rfc6455 and https://html.spec.whatwg.org/multipage/web-sockets.html#network) on not supporting HTTP header.

Talking about `connectionParams`, it is also discussed at https://github.com/apollographql/apollo-client/issues/3967#issuecomment-435808419 by @pkosiec that
> It won't work as WebSocket API in browsers doesn't support setting custom headers, apart from the value of `Sec-Websocket-Protocol` header.
https://stackoverflow.com/questions/4361173/http-headers-in-websockets-client-api/41521871

Anyway, with the change introduced in this PR, I hope i'm not breaking `connectionParams`/`initPayload` 🤔

packages/graphql/lib/src/link/web_socket/link_web_socket.dart
```diff
@@ -23,15 +22,13 @@ class WebSocketLink extends Link {
   /// Creates a new [WebSocketLink] instance with the specified config.
   WebSocketLink(
       {@required this.url,
-      this.headers,
```
micimize: How about, instead of removing it entirely, we:
* mark the header argument as as deprecated
* throw an error `if (dart.library.html)`
* warn when it's used `if (dart.library.io)` similar to the legacy payload

```dart
if (headers != null) {
  assert(
    !dart.library.html,
    'Cannot set websocket headers with dart:html websockets. '
    'If these are for authentication, another approach must be used, '
    'such as initPayload.',
  );
  print(
    'WARNING: Using direct websocket headers which will be removed soon, '
    'as it is incompatable with dart:html. '
    'If you need this direct header access, '
    'please comment on this PR with details on your usecase: '
    'https://github.com/zino-app/graphql-flutter/pull/323',
  );
  concatHeaders.addAll(headers);
}
```

truongsinh: agree, will make appropriate change.

truongsinh: please review the latest change, I have this assertion in constructor instead.

micimize:
@truongsinh So, after looking at / thinking about it more, I realized accepting the headers in the `dart:io` case is kinda useless given we're now using your [universal websocket package](https://github.com/truongsinh/dart-websocket).

So, instead of trying to maintain a bunch of legacy routes throughout the websocket api, I decided to just split out all the old behavior into the `legacy_socket_api`. This includes the `legacyInitPayload`

.circleci/config.yml
```diff
@@ -102,6 +102,8 @@ jobs:
             cd packages/graphql
             pub global activate coverage
             pub get
+            # warm-up dart2js, this is a dirty workaround
+            dart2js test/websocket_test.dart
```
HofmannZ: Could you explain why this ‘dirty workaround‘ is necessary?

micimize:
Hmm - well, idk why the CI is failing now 😓 . Anyways, @truongsinh take a look at my refactor - we'll need to add docs pointing to the legacy stuff, but all the websocket docs are out of date and lacking anyways so we can do that separately

.circleci/config.yml
micimize: IDK, but it might not be. I added it back when CI started failing because I thought it might have been the cause.

truongsinh: in the last ~10 CI running, the browser test always fails to compile the first file 
```
00:04 +0 -1: compiling test/multipart_upload_test.dart [E]                                                                                                                                             
  Failed to load "test/multipart_upload_test.dart": dart2js failed.
```
and the failing file is also random. I'm trying to figure out, but not yet succeed. If I can't figure in time, I suggest we temporarily disable `pub run test test -p chromium`

truongsinh:
@micimize I'm fine with the refactoring commit, in the following commits, I move 4 test cases so that it can be tested on both VM and browser, and only leave 2 which fails on browser.

Issue #297, opened 2019-05-24
refactor: Clean up caching submodule


micimize:
- [x] **refactor**: The cache submodule is kinda a huge mess of hasty inheritance
- [x] **api tuning**
  - [x] we should have a more descriptive top-level cache export like` GraphQLCache`
  - [x] Consider implementing the same interface as [quiver.cache](https://pub.dev/documentation/quiver/latest/quiver.cache/quiver.cache-library.html)
- [ ] **standalone** Almost all of the cache logic could be leveraged by other modules (such as angel) 
  it might be worth making a standalone module, but it could at least be more accessible as a standalone

micimize:
the cache could easily be made stand alone but am not entirely sure what it gets us. If another lib is interested in using it we can def do that in the future though

PR #325, opened 2019-06-02
feat(client): cache is supported in web/browser platform

truongsinh:
This is an ongoing effort to make this package compatible to web/browser #295.

In web/browser platform, Cache is supported by `localStorage`.

The naming `InMemoryCache` is also debatable, as both `io` and `html`
implementation of this class is based on "persistent storage", not at all
"in memory".

There are refactor works to do to consolidate `in_memory_*` files.

### Breaking changes

`InMemoryCache` no longer accept `storageProvider`,
however this param has always been using for testing only.


micimize:
This is related to #297 - I really think caching is worth breaking into a separate module https://pub.dev/packages/caches

anyways @truongsinh I'm guessing rebasing will resolve these conflicts and then I can re-approve and merge

truongsinh:
@micimize please be aware that I have a new commit, now the test is running in headless Chromium as well

PR #330, opened 2019-06-07
code coverage badge per package

truongsinh:
ci: code coverage badge per package


micimize:
@truongsinh can you set up automatic commit signing?

PR #342, opened 2019-06-15
feat: added headers parameter to access response headers

aagarwal1012:
### Breaking changes

- I need to get the value of cookie from the response headers but FetchResult and QueryResult don't have the headers parameter. 

#### Fixes / Enhancements

- Added headers parameter in FetchResult and QueryResult.


micimize:
Closing in favor of considering the approach outlined in #371

PR #350, opened 2019-06-28
Add FetchMore Feature

mainawycliffe:
This adds the fetch more feature for queries within the Query Widget. It is an early version, so am open to ideas and suggestions, especially on some use cases i may not have foreseen.

### Breaking changes

This breaks the current query builder method, by exposing a fetchMore method which can be called within the builder to fetch more documents.

So instead of 
```dart
...
builder: (QueryResult result, {refetch}) {
  ...
}
...
```

it will be:

```dart
...
builder: (QueryResult result, {refetch, FetchMore fetchMore}) {
  ...
}
...
```
#### Fixes / Enhancements

- Attempts to solve [this](https://github.com/zino-app/graphql-flutter/issues/144).

#### Docs

For the time being, i have update the `graphql-flutter` [example](https://github.com/mainawycliffe/graphql-flutter/blob/fetchmore/packages/graphql_flutter/example/lib/fetchmore/main.dart), to add a third example demonstrating this feature in action. Once we have addressed all concerns that may arise , i will update docs and tests accordingly.

### DIFF
--- a/examples/starwars/lib/client_provider.dart
+++ b/examples/starwars/lib/client_provider.dart
@@ -20,7 +20,7 @@ ValueNotifier<GraphQLClient> clientFor({
   @required String uri,
   String subscriptionUri,
 }) {
-  Link link = HttpLink(uri: uri) as Link;
+  Link link = HttpLink(uri: uri);
   if (subscriptionUri != null) {
     final WebSocketLink websocketLink = WebSocketLink(
       url: subscriptionUri,
--- a/examples/starwars/lib/episode/hero_query.dart
+++ b/examples/starwars/lib/episode/hero_query.dart
@@ -31,7 +31,11 @@ class HeroForEpisode extends StatelessWidget {
           'ep': episodeToJson(episode),
         },
       ),
-      builder: (QueryResult result, {BoolCallback refetch}) {
+      builder: (
+        QueryResult result, {
+        BoolCallback refetch,
+        FetchMore fetchMore,
+      }) {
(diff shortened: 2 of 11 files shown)


lucasavila00:
It doesn't work while using normalized cache.
I don't know why but its probably because of the way it's lazily stored.

I thought about a reducer function iterate over all responses and return the data to render.
This way there isn't issues with the cache, I think.

mainawycliffe:
@lucasavila00 let me see what I can do. I will do some testing over the weekend, i would like to see this cover most of the use cases if not all. And thanks for your example, i will see if i can borrow some elements from it.

lucasavila00:
@mainawycliffe I'm developing a feature right now and using the code I wrote.
I'm upgrading it to be able to deal with edge cases, if you'd like you can check the live version here https://github.com/lucasavila00/graphql-flutter/blob/fork2/packages/graphql_flutter/lib/src/widgets/query_fm.dart

Maybe it helps you :)

chimon2000:
Very excited to see this feature land!

jensdev:
Any update on when this feature will land?

micimize:
@jensdev sorry for the delay - I've been pretty swamped. Will try and get to this and some other reviews sometime this week

mainawycliffe:
Just went over it, looks good to me. Unfortunately, I am not authorized to merge, you can merge it when you are able to.

eusdima:
@mainawycliffe I merged. I will ask Zino about merging rights.

PR #352, opened 2019-06-29
Prevent query from running again if not necessary

lucasavila00:
Whenever I push a new screen on top of another one that has a query the bottom one also updates it's content.
However, when I pop to this screen the query is not rerun (I don't know if this was wanted).
Anyway, I think the old implementation is really bad because if I have 10 screens on top on one another (user opens a post then opens creator's page then open another post) then I will have 10 queries running all the time.

I don't know if it was intentional to depend on another part of the library to change (like if you changed the client link) but then it would have to be done in another way.

I've been using it like this and haven't found any problems so far.

### Breaking changes

- Broke ... because ... .

#### Fixes / Enhancements

- Fixed ... was ... .
- Added ... .
- Updated ... .

#### Docs

- Added ... .
- Updated ... .

### DIFF
--- a/packages/graphql_flutter/lib/src/widgets/query.dart
+++ b/packages/graphql_flutter/lib/src/widgets/query.dart
@@ -61,7 +61,10 @@ class QueryState extends State<Query> {
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
-    _initQuery();
+    if (observableQuery == null ||
+        !observableQuery.options.areEqualTo(_options)) {
+      _initQuery();
+    }
   }
 
   @override


lucasavila00:
If you can wait I can submit tests for what I'm saying.

micimize:
I actually removed something similar in #110 because I thought `didChangeDependencies` would only be called when the `GraphQLProvider` (and thus `client`) was replaced, but apparently [all widgets get `didChangeDependencies` called when pushing a route](https://github.com/flutter/flutter/issues/18774#issuecomment-503593201)

I guess we need to save the a client reference in the state, and add a comparison to the conditional there

serendipity1004:
What's the hold? Is this ready yet?

micimize:
@serendipity1004 Sorry, my last comment wasn't clear enough.

We need to add a reference to the `Client` as well as corresponding `didChangeDependencies` checks in `Query` and `Mutation`. Otherwise the widget observables will never reattach to a new `Client`, such as in the hot-reload scenario I encountered (https://github.com/zino-app/graphql-flutter/pull/110#issuecomment-427435474)

HofmannZ:
Please `reword` your commit messages to align with our [contributing guidlines](https://github.com/zino-app/graphql-flutter/blob/master/CONTRIBUTING.md).

fredericp:
I agree with @micimize.  We need to store the current client in the widget and check if the client has changed in didChangeDependencies. We also have to fix Mutation and Subscription. 

I have a fix in https://github.com/fredericp/graphql-flutter/commit/7c52fe7052f06f76193b600e5b8cb609f73e12f0 maybe we can update the PR or make another one.

PR #359, opened 2019-07-16
fix(client): error handling

truongsinh:
Fixing https://github.com/zino-app/graphql-flutter/issues/355 that some errors/exceptions that is not related to GraphQL are still coerced and have no useful information.

### Breaking changes

- Should not be a breaking change, because previously, we already expect some exceptions/errors to be thrown directly

### DIFF
--- a/packages/graphql/test/graphql_client_test.dart
+++ b/packages/graphql/test/graphql_client_test.dart
@@ -128,7 +128,32 @@ void main() {
         expect(nodes[2]['viewerHasStarred'], false);
         return;
       });
-//    test('failed query because of network', {});
+
+      test('failed query because of an exception with null string', () async {
+        final e = Exception();
+        when(mockHttpClient.send(any)).thenAnswer((_) async {
+          throw e;
+        });
+
+        final Future<QueryResult> r = graphQLClientClient
+            .query(WatchQueryOptions(document: readRepositories));
+
+        await expectLater(r, throwsA(e));
+        return;
+      });
+
+      test('failed query because of an exception with empty string', () async {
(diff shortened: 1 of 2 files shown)


packages/graphql/test/graphql_client_test.dart
```diff
@@ -128,7 +128,30 @@ void main() {
...
         return;
       });
-//    test('failed query because of network', {});
+
+      test('failed query because of an exception with null string', () async {
+        final e = Exception();
+        when(mockHttpClient.send(any)).thenAnswer((_) async {
+          throw e;
+        });
+        final Future<QueryResult> r = graphQLClientClient
```
HofmannZ: Nit: add a new line between the mock and the test execution. This will make it easier to read.

packages/graphql/test/graphql_client_test.dart
```diff
@@ -128,7 +128,30 @@ void main() {
...
+
+        await expectLater(r, throwsA(e));
+        return;
+      });
+
+      test('failed query because of an exception with empty string', () async {
+        final e = Exception('');
+        when(mockHttpClient.send(any)).thenAnswer((_) async {
+          throw e;
+        });
```
HofmannZ: Nit: add a new line between the mock and the test execution. This will make it easier to read.

packages/graphql/test/graphql_client_test.dart
truongsinh: resolved

micimize:
@truongsinh does this also need to be PRed against `beta` as well?

PR #370, opened 2019-08-02
Pass the storage path to the InMemoryCache as a parameter

mateusfsilva:
Allow to passing the storage path as a parameter of the InMemoryCache initializer.

#### Fixes / Enhancements

- Fixed a problem on our tests where the dependency of the "path_provider" package throws an error.
- Added the possibility to save the cache to a directory other than the documents directory.


micimize:
switched target to `beta`, other than that LGTM

PR #393, opened 2019-08-26
Library-level Exceptions, result.exception

### LINKED ISSUE #319: custom library-level exceptions
As pointed out in #177
>  we could solve this by creating our own kind of exceptions so that we can handle them cleaner internally.

It would also make it easier / possible for users to have different logic for different errors, i.e. ignore network related issues when using the cache. 

Worth noting that without something like #201, it will be easy for a user making heavy use of the offline queue to shoot themselves in the foot by pretending mutations succeeded when they didn't and will never run again.

### LINKED ISSUE #380: SocketException is crashing application
**Describe the bug**
I cannot handle the SocketException error.

**To Reproduce**
1. Go to https://github.com/ryankauk/graphqlFlutterSocketException
2. Install and run flutter application. (I'm using VsCode Start Debugging using Flutter extension)
3. You'll see the error thrown

**Expected behavior**
The error should be at the minimum be able to handled in try catch statement. Stated by the package it should be added to graphql errors.

**Screenshots**
When using try catch
Screen Shot 2019-08-13 at 2 30 48 PM

Full error without try catch
Screen Shot 2019-08-13 at 2 30 17 PM

 - OS: macOS Mojave 
 - Version 10.14.4

micimize:
closes #319
closes #380

### Breaking changes

- Broke `QueryResult.errors` in order to wrap it in `OperationException`

#### Fixes / Enhancements

* introduced distinction between `GraphqlErrors` and `ClientExceptions`
* make `SocketExceptions` intelligible via `NetworkException`
* `OperationException` `QueryResult.exception` that supports both
  * `clientException` such as `NetworkException` and `CacheMissException`
  * and `graphqlErrors`
* added paginating tab to graphql example

#### Docs
Added to readme:

##### Exceptions

If there were problems encountered during a query or mutation, the `QueryResult` will have an `OperationException` in the `exception` field:

```dart
class OperationException implements Exception {
  /// Any graphql errors returned from the operation
  List<GraphQLError> graphqlErrors = [];

  /// Errors encountered during execution such as network or cache errors
  ClientException clientException;
}
```

Example usage:

```dart
if (result.hasException) {
    if (result.exception.clientException is NetworkException) {
        // handle network issues, maybe
    }
    return Text(result.exception.toString())
}
```


micimize:
@HofmannZ please take a [quick look](https://github.com/zino-app/graphql-flutter/pull/393/files?file-filters%5B%5D=.dart) at this, as I think it's a significant api design change.

examples/starwars/lib/episode/hero_query.dart
```diff
@@ -36,15 +36,15 @@ class HeroForEpisode extends StatelessWidget {
...
-        }
-
         if (result.loading) {
-          return Center(
-            child: const CircularProgressIndicator(),
+          return const Center(
+            child: CircularProgressIndicator(),
           );
         }
+
```
HofmannZ: Would always check for errors before checking for loading. To abound infinite loading screens when actually an error had occurred.

examples/starwars/lib/reviews/force_rebroadcast.dart
```diff
@@ -0,0 +1,31 @@
...
+import 'package:graphql_flutter/graphql_flutter.dart';
+
+class ForceRebroadcast extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return Mutation(
+      options: MutationOptions(
+        document: r'''
+        mutation Rebroadcast {
+        }
```
HofmannZ: Is this valid query language?

packages/graphql/lib/src/core/query_manager.dart
```diff
@@ -36,12 +36,6 @@ class QueryManager {
   Map<String, ObservableQuery> queries = <String, ObservableQuery>{};
 
   ObservableQuery watchQuery(WatchQueryOptions options) {
-    if (options.document == null) {
```
HofmannZ: Can we safely drop this?

packages/graphql/lib/src/core/query_manager.dart
```diff
@@ -120,6 +114,9 @@ class QueryManager {
         );
       }
 
+      // TODO this should never happen right?
+      // if there is an error reaching the server, there should be an error thrown,
+      // if not, data should not be null.
       if (fetchResult.data == null &&
```
HofmannZ: Yeah that’s right!

packages/graphql/lib/src/core/query_options.dart
```diff
@@ -92,8 +92,8 @@ class QueryOptions extends BaseOptions {
   QueryOptions({
     @required String document,
     Map<String, dynamic> variables,
-    FetchPolicy fetchPolicy = FetchPolicy.cacheFirst,
-    ErrorPolicy errorPolicy = ErrorPolicy.none,
+    FetchPolicy fetchPolicy,
```
HofmannZ: Why take out these defaults?

packages/graphql/lib/src/core/raw_operation_data.dart
```diff
@@ -13,7 +13,8 @@ class RawOperationData {
     @required this.document,
     Map<String, dynamic> variables,
     String operationName,
-  })  : _operationName = operationName,
+  })  : assert(document != null),
```
HofmannZ: At the comment earlier we had a readable error. No we have an failing assertion on the `RawOperation` that might be a load harder to debug for someone new to the library.

packages/graphql/lib/src/exceptions/_base_exceptions.dart
```diff
@@ -0,0 +1,46 @@
+abstract class ClientException implements Exception {
+  String get message;
+}
+
+//
+// Cache exceptions
```
HofmannZ: Nice to have all the errors together! 👍🏼

examples/starwars/lib/reviews/force_rebroadcast.dart
micimize: no - I'm removing it as it was just a utility for testing some pagination issues.

packages/graphql/lib/src/core/query_manager.dart
micimize: yes, because of `assert(document != null)` in `RawOperationData`

packages/graphql/lib/src/core/query_options.dart
micimize: Defaults are now configurable in `GraphQLClient.defaultPolicies` (#398), so they need to be `null` to be overridden later. So this is actually more of a fix 😅

packages/graphql/lib/src/core/raw_operation_data.dart
micimize: added a the message back in on failed assertion

micimize:
@HofmannZ changes made and questions answered 👆

HofmannZ:
Should we bundle this with #414, so we only ship one breaking change?

mainawycliffe:
> Should we bundle this with #414, so we only ship one breaking change?

I think this is better, so we could have a milestone for the next major version and see what major API changes we could fit it in.

micimize:
> Should we bundle this with #414, so we only ship one breaking change?

I agree, but only in terms of doing so when we exit `beta` (merge into `master`)

HofmannZ:
@micimize, @mainawycliffe Agreed, I’ve pushed some fixed to the pipeline, so we should see releases more often.

Issue #364, opened 2019-07-18
Subscriptions Reconnect solution


rullyalves:
I noticed that graphql_flutter does not have a functional reconnection of subscriptions when the internet connection is interrupted, I created an implementation to take care of that and would like to integrate with the current package, my implementation has already been tested and is fully functional

mainawycliffe:
Fee free to create a PR against beta, follow contributions guidelines [here](https://github.com/zino-app/graphql-flutter/blob/master/CONTRIBUTING.md). Make sure to include tests and if possible update documentation where necessary.

bsr203:
@rullyalves would like to see your implementation. Thanks.

kdkul:
@mainawycliffe @rullyalves any idea when this solution will be available? I am looking for this solution to implement in one of my project. @rullyalves thanks for valuable solution.

mainawycliffe:
@rullyalves are you able to create a PR for this?

PR #403, opened 2019-09-02
Reconnect websocket subscription

MSchmack:
Updating the subscription data when the user goes from offline to an online state with an ongoing subscription.

#### Fixes / Enhancements

Solution for issue #364 
Checks when to reconnect with the connectivity plugin. Calls _initSubscription when internet should be available again.

> Note that on Android, this does not guarantee connection to Internet. For instance, the app might > have wifi access but it might be a VPN or a hotel WiFi with no access.
Uses a nslookup on connectivitychange for android as solution to the shortcoming of the connectivity plugin.

It might be wise to make the android nslookup optional for users that want to reduce traffic, even if minimal.

#### Docs

No document change needed as this functionality is already implied with the flag autoReconnect.

### DIFF
--- a/packages/graphql_flutter/example/pubspec.yaml
+++ b/packages/graphql_flutter/example/pubspec.yaml
@@ -6,6 +6,7 @@ dependencies:
     sdk: flutter
   cupertino_icons: ^0.1.2
   rxdart: ^0.22.0
+  connectivity: ^0.4.4
   graphql_flutter:
     path: ..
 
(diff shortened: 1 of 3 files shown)


HofmannZ:
@MSchmack Still missing tests.

MSchmack:
@HofmannZ There are no tests in the graphql_flutter project for the different widgets so none have been expanded. Do you test the widgets at a different location?

https://github.com/zino-app/graphql-flutter/tree/master/packages/graphql_flutter/test

HofmannZ:
@MSchmack My apologies I thought this was for the subscription client. LGTM.

smkhalsa:
@HofmannZ shouldn't this be added to the client rather than the widget directly?

ryankauk:
Is there an estimated time when this would reach stable?

PR #408, opened 2019-09-06
refetchQuery should return a future

smkhalsa:


mainawycliffe:
Why should refetchQuery return a future? Please articulate the reasons clearly and IMO this discussion should happen before a PR, by raising an issue especially since its a breaking change.

smkhalsa:
@mainawycliffe yes, apologies.

Just as fetchQuery returns a Future, I believe that refetchQuery should as well. In my use case, I am calling refetchQuery from my store, and I need to know when the refetch completes before showing a user message.

budde377:
It would be nice if this was propagated all the way out to the `Query` widget.

serendipity1004:
@budde377 yes. is it currently not possible to use refetch with RefreshIndicator?

budde377:
I might be missing something, but refresh indicaror requires uses a future
to indicate when the data has been refreshed. While you can of course just
return immediately with refetch, this is far from ideal.

serendipity1004:
@budde377 I agree that this is necessary. Currently just providing refetch to RefreshIndicator returns error because it does not return a future.

mainawycliffe:
I understand why it needs to be a future, but as the PR currently is, it is breaking other parts of the library as I highlighted when requesting changes. I was giving @smkhalsa time to make changes I requested before I could approve it. I guess since I haven't heard from him, I will make the changes before approving the PR.

budde377:
@mainawycliffe If you'd like, I can give it a go. If so, would you prefer that we modify the signature of the current `refetchQuery` methods or add another method, e.g. `refetchQueryAsync`, which returns a `Future`?

mainawycliffe:
@budde377 Go ahead. I think modifying the current `refectchQuery` method is better, no need to bloat the API unnecessarily. Just make sure you update the parts of the library relying currently on the functionality by adding await in front of the async `refectchQuery` method.

smkhalsa:
@mainawycliffe apologies for not responding to this sooner. I've been traveling internationally.

I see that you requested changes, but for some reason the actual changes requested aren't showing up for me.

In any case, @budde377 are you working on this now, or should I give it another pass?

budde377:
@smkhalsa I haven't started yet so please go ahead 😊 As I mentioned previously, it would be extremely useful to have this propagated all the way out to the `Query` widget.

smkhalsa:
Closing in favor of #425.

PR #414, opened 2019-09-14
[Breaking] WIP: Use AST-based Link from `package:gql`

klavs:
This PR is just to show what it would take to use AST documents instead of String documents.

Since the document is already parsed in order to get the `operationName` this change only introduces an additional action per request – printing the AST to GraphQL source.

`package:gql/dart` has dart code generation so with some additional work, the document parsing can be moved from run-time to build-time.

`package:gql/ast` has means of transforming the AST, therefor adding `__typename` to selection sets gets one huge step closer.

<hr>

**_Update_**

_Disclosure: so far I am the only contributor to `package:gql`, but contributions are welcome_

I've extended this PR to use `package:gql/link` as the abstraction for `Link`. It's based on AST documents instead of string documents and provides a type-aware `Context`. It has a dependency on `package:gql/execution`, so I'm refactoring to use the types from that library.

This is a breaking change.


HofmannZ:
@klavs Thanks for the PR, please rebase on beta, we had to fix the coverage collection in the pipeline.

HofmannZ:
Also this would be a breaking change, and should be released as such

HofmannZ:
Please check your commit messages to align with our [contributing guidlines](https://github.com/zino-app/graphql-flutter/blob/master/CONTRIBUTING.md).

klavs:
@HofmannZ, I'll fix the messages later.

Since this is a breaking change I do not want to rush it. I was actually more looking for feedback about this whole idea. To me it makes sense to align the `Link` to what Apollo does.

HofmannZ:
@klavs The idea is great! We want to make it as easy a possible to get started when comping for existing GraphQL clients. Apollo is currently the biggest out there, so it makes sense aligning our library with it.

klavs:
I'm now working on a compatible change which would allow to use AST documents without breaking the all the interfaces I've broken with this PR.

I think that should be a good transitional step. But I think, it would be good to open up a branch for the next major version.

Also I'm working on improving the `package:gql/link` to make the transition to AST-based links smoother.

HofmannZ:
@klavs agreed let's do that in the alpha branch

PR #429, opened 2019-10-05
AST-based operations

klavs:
Allow creating operations out of AST.
This is a non-breaking transitional to enable AST-based links, local state resolvers.

This change add as slight performance reduction because documents must be printed to string before sending them over the network.
The remedy for that is building AST at build-time. Currently that is possible using `package:gql_code_gen`.

### DIFF
--- a/packages/graphql/lib/src/core/query_options.dart
+++ b/packages/graphql/lib/src/core/query_options.dart
@@ -1,3 +1,5 @@
+import 'package:gql/ast.dart';
+import 'package:gql/language.dart';
 import 'package:meta/meta.dart';
 
 import 'package:graphql/src/utilities/helpers.dart';
@@ -47,6 +49,7 @@ class Policies {
 
   /// Specifies the [ErrorPolicy] to be used.
   ErrorPolicy error;
+
   Policies({
     this.fetch,
     this.error,
@@ -67,12 +70,18 @@ class Policies {
 /// Base options.
 class BaseOptions extends RawOperationData {
   BaseOptions({
-    @required String document,
+    @Deprecated('The "document" option has been deprecated, use "documentNode" instead')
+        String document,
+    DocumentNode documentNode,
     Map<String, dynamic> variables,
     this.policies,
     this.context,
     this.optimisticResult,
-  }) : super(document: document, variables: variables);
+  }) : super(
+          document: document,
+          documentNode: documentNode,
+          variables: variables,
+        );
 
   /// An optimistic result to eagerly add to the operation stream
   Object optimisticResult;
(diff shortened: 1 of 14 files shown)


klavs:
@HofmannZ, here's the non-breaking change I was referring to in https://github.com/zino-app/graphql-flutter/pull/414

packages/graphql/lib/src/core/query_options.dart
```diff
@@ -67,12 +69,17 @@ class Policies {
 /// Base options.
 class BaseOptions extends RawOperationData {
   BaseOptions({
-    @required String document,
+    @Deprecated('Use documentNode instead') String document,
```
HofmannZ: Good to first deprecated support for `document`. Maybe rephrase the message to:

```
'The "document" option has been deprecated, use "documentNode" instead'
```

packages/graphql/lib/src/core/query_options.dart
```diff
@@ -194,12 +207,25 @@ typedef dynamic UpdateQuery(
...
+    this.documentNode,
     this.variables = const <String, dynamic>{},
     @required this.updateQuery,
-  }) : assert(updateQuery != null);
+  })  : assert((document != null && documentNode == null) ||
+            (document == null && documentNode != null)),
+        assert(updateQuery != null);
+
+  DocumentNode documentNode;
```
HofmannZ: Why not use getters and setters?

packages/graphql/lib/src/core/raw_operation_data.dart
```diff
@@ -31,7 +45,7 @@ class RawOperationData {
   String get operationName {
     // XXX there is a bug in the `graphql_parser` package, where this result might be
     // null event though the operation name is present in the document
-    _operationName ??= getOperationName(document);
+    _operationName ??= getLastOperationName(documentNode);
```
HofmannZ: Can we test if this bug is fixed?

packages/graphql/lib/src/core/query_options.dart
klavs: OK, added a more verbose deprecation message

packages/graphql/lib/src/core/query_options.dart
klavs: Added the missing initialisation to the constructor

packages/graphql/lib/src/core/raw_operation_data.dart
klavs: Removing the comment. It's not too helpful in describing the exact conditions of the bug.

If any bug in parser appears, we'll file a bug to `package:gql` and we'll fix it there.

klavs:
@HofmannZ I addressed the feedback and signed the commits.

I think the CI fails incorrectly – we should be able to deprecate our own API and use it while it is deprecated.

cc: @mainawycliffe

klavs:
added info about AST documents to the readme.

Note that I'm assuming the version `2.2.0`

lookfirst:
This is awesome!

PR #430, opened 2019-10-05
ci: allow coverage to drop by up to 1%

HofmannZ:
This PR allows coverage to drop by up to 1% in a PR before marking it failed.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,7 +1,15 @@
 flags:
-  graphql_flutter:
-    paths:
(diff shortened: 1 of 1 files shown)


mainawycliffe:
LGTM. Can we go over the library and see which areas we would add more test on.

HofmannZ:
@mainawycliffe Yeah, something I discussing with @eusdima. Really time-consuming, but really rewarding..

PR #431, opened 2019-10-05
ci: bump the `semantic-release` version

HofmannZ:
Nightly releases are currently broken because `semantic-release` is on a broken version. This PR bumps the version to fix the issue.

This PR also remove the `sync_with_beta` when running on the beta branch. Merging into your own branch does not work of course.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,7 +122,7 @@ jobs:
       - checkout
       - run:
           name: Generate release tags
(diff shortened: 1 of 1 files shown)


mainawycliffe:
I have also notice that master is not getting merged into beta.

HofmannZ:
@mainawycliffe yeah that sttep fires after a successful release, so did not trigger because of this issue.

PR #433, opened 2019-10-06
test(client): add tests for `operationName` getter

klavs:


packages/graphql/test/core/raw_operation_data_test.dart
```diff
@@ -0,0 +1,125 @@
...
+
+      test('subscription with implicit name', () {
+        final opData = RawOperationData(
+          document: 'subscription Operation {}',
+        );
+
+        expect(opData.operationName, 'Operation');
+      });
+    });
+    group('multiple operations', () {
```
HofmannZ: Just nitpicking, I know, but a white line between groups would be nice!

Issue #365, opened 2019-07-22
Add an onError callback to Mutation component


kasonde:
**Describe the bug**
When using a mutation, I'm unable to tell whether there are mutation has errors or not in the onComplete callback of the mutation component. i.e

```
Mutation(
...,
onComplete: (result){
    // result is a json object. result.hasErrors doesn't work anymore
})
```

mainawycliffe:
Yes, as currently  implemented, it returns the data only and not a `QueryResult` class which would expose the error. I am not sure why this was implemented this, so i need to do some more research.

mainawycliffe:
After some initial research, the API works as expected returning data on complete, but it is missing `onError` which is a callback for catching when errors occur during mutation.

kasonde:
@mainawycliffe  any idea of the progress of this? Do you have any workarounds?

mainawycliffe:
@kasonde you can access errors on the `onUpdate` callback, which returns a `QueryResult` Object, check the example given [here](https://github.com/zino-app/graphql-flutter/blob/16fb1d507c44c24c0cf7da56dac304b956a95643/packages/graphql_flutter/example/lib/graphql_widget/main.dart#L211).

```dart
...
update: (Cache cache, QueryResult result) {
  if (result.hasErrors) {
    print(result.errors);
  } 
  ...
}
...
```

mainawycliffe:
@micimize should we add an `OnError` method to be called in case of error, just like the `onCompleted` callback is called in case of data?

micimize:
@mainawycliffe yes - appropriating this issue for that.

I think we should take the opportunity to implement an analogue to [`ApolloError`](https://github.com/apollographql/apollo-client/blob/7eaf4132cd2cd6244260777799406aaa03fcf377/packages/apollo-client/src/errors/ApolloError.ts#L33) as a step towards #319 (making it a breaking api change) and implement `onError(OurCustomErrorClass)`, bringing us closer to [apollo's `Mutation` props](https://www.apollographql.com/docs/react/essentials/mutations/#props)

Deciding on what to name the error is tough - maybe `GraphQLError` or `GraphQLErrorContainer`?

mainawycliffe:
I will get on this over the weekend.

mainawycliffe:
Closed by #435

Issue #278, opened 2019-05-19
Move "update" function from flutter component to client


smkhalsa:
In line with [Apollo](https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/core/QueryManager.ts#L134), we should add an `update` function as well as a `refetchQueries` array within the `MutationOptions` of QueryManager's `mutate` function,

micimize:
`update` logic exists but needs to be moved from `graphql_flutter`
We should also add `refetchQueries` - but only the part they don't [want to deprecate](https://github.com/apollographql/react-apollo/blob/master/src/Mutation.tsx#L205-L213)

Right now all "safe" queries are re-resolved from the cache on mutation, so there's that 😅

smkhalsa:
@mainawycliffe Do you have any guidance on this?

mainawycliffe:
@smkhalsa At first I didn't understand what you meant, but now I see the difference between our own implementation and React Apollo implementation. Our Update method is outside mutation options, which makes it only accessible to users using `graphql_flutter` and not regular `graphql`. I think it is possible to move the update the method, to allow users of `graphql` to use the option too. @micimize what do you think?

smkhalsa:
Thanks @mainawycliffe. My use case requires that I initiate a mutation from my store in response to state changes as opposed to something directly initiated by a user in the UI. As such, I'm using the `graphql` library directly, not `graphql_flutter`.

micimize:
Yeah we should move the callback logic involved into `graphql`, and just pass through the current update callback. This applies to pretty much all logic - `graphql_flutter` shouldn't really have much logic in it

smkhalsa:
@mainawycliffe it sounds like this is a change we want to make. Any further thoughts?

mainawycliffe:
@smkhalsa yea, this is a change we should make, the question is the best approach to achieve this. It will definitely be a breaking change, but if we move the `update` call back alone, I think we could introduce some API inconsistencies, the best approach, I believe, is to move all the callbacks to `MutationsOptions` - `onError`, `onComplete`, `update` which is more consistent with [Apollo React](https://www.apollographql.com/docs/react/data/mutations/#options).

micimize:
@mainawycliffe yeah that's what I think as well - all functionality should be in `graphql` and merely exposed in an idiomatic flutter way in `graphql_flutter`

mainawycliffe:
I will work on this and bundle it with the #435

[...]

PR #435, opened 2019-10-06
Move Mutation Callbacks to MutationOptions

mainawycliffe:
This PR moves Mutation Callbacks to under `MutationOptions` so that more logic is in the GraphQL Client instead of widget side and to be more in line with apollo, where these callbacks are directly under MutationOptions. This also attempts to expose some of those callbacks so that they can be used by developers using the graphql client (#278). 

On top of that, I have added a `onError` callback to compliment both the `update` and `onComplete` callbacks just like in apollo. Open to the idea of improvements to this PR and feedback from those who will test it.

#### Breaking Changes

This going to be a breaking change as the Mutation Callbacks needs to be called directly under MutationOptions instead of the Widget Builder.

before:

```dart
Mutation(
  options: MutationOptions(
    document: addStar, // this is the mutation string you just created
  ),
  builder: (
    RunMutation runMutation,
    QueryResult result,
  ) {
    return FloatingActionButton(
      onPressed: () => runMutation({
        'starrableId': <A_STARTABLE_REPOSITORY_ID>,
      }),
      tooltip: 'Star',
      child: Icon(Icons.star),
    );
  },
  // you can update the cache based on results
  update: (Cache cache, QueryResult result) {
    return cache;
  },
  // or do something with the result.data on completion
  onCompleted: (dynamic resultData) {
    print(resultData);
  },
);
```

after:

```dart
Mutation(
  options: MutationOptions(
    document: addStar,
     // you can update the cache based on results
    update: (Cache cache, QueryResult result) {
      return cache;
    },
    // or do something with the result.data on completion
    onCompleted: (dynamic resultData) {
      print(resultData);
    },
  ),
  builder: (
    RunMutation runMutation,
    QueryResult result,
  ) {
    return FloatingActionButton(
      onPressed: () => runMutation({
        'starrableId': <A_STARTABLE_REPOSITORY_ID>,
      }),
      tooltip: 'Star',
      child: Icon(Icons.star),
    );
  },

);
```

#### Fixes / Enhancements

- This fixes issues #365 and #278


packages/graphql/lib/src/core/query_options.dart
```diff
@@ -118,12 +124,106 @@ class MutationOptions extends BaseOptions {
     FetchPolicy fetchPolicy,
     ErrorPolicy errorPolicy,
     Map<String, dynamic> context,
```
smkhalsa:
```suggestion
    Map<String, dynamic> context,
    Object optimisticResult,
```

packages/graphql/lib/src/core/query_options.dart
```diff
@@ -118,12 +124,106 @@ class MutationOptions extends BaseOptions {
...
+    this.onCompleted,
+    this.update,
+    this.onError,
   }) : super(
           policies: Policies(fetch: fetchPolicy, error: errorPolicy),
           document: document,
           variables: variables,
           context: context,
```
smkhalsa:
```suggestion
          context: context,
          optimisticResult: optimisticResult,
```

packages/graphql/lib/src/core/query_options.dart
mainawycliffe: I think this needs to be more thought out implementation and should be a separate issue. It is not as easy as moving existing APIs from one package to another.

micimize:
Rebased and added a fetchmore regression and testing fixes from #429

@HofmannZ This is good to go other than CI hanging

PR #436, opened 2019-10-07
Revert "Library-level Exceptions, result.exception"

klavs:
Reverts zino-app/graphql-flutter#393

@HofmannZ There are multiple issues with this merge:

- it includes unrelated ⚠️BINARY⚠️ files (I think `.gitignore` should be updated)
- it was not rebased so it made the tests for `HttpLink` (merged in earlier) to fail


micimize:
@klavs we can go ignore binaries, etc. And fix tests, reverting doesn't really yield an advantage here

micimize:
I think we should close this in favor of #441, but will leave open in case there were other issues @klavs had with #393

klavs:
We can close this, @micimize

I was hoping for a quick resolution, but this PR does not matter anymore

Issue #440, opened 2019-10-13
Add ErrorLink


smkhalsa:
Something comparable to Apollo's ErrorLink

bsr203:
Similar request https://github.com/zino-app/graphql-flutter/issues/419

I too really like to have this feature to catch Network error and to handle appropriately outside components. I see an approach in this repo https://github.com/ygmpkk/flutter_graphql_backenderrorlink 
but, as far as I see, ClientExceptions from http package is not able to capture by this package and may need to implement at core.

jayjun:
@smkhalsa @bsr203 I sent a pull request (#445) to add `ErrorLink`.

bsr203:
@jayjun thank you so much for your work on it. This was much needed feature for me. I didn't test it, but there has been lot changes around splitting network error etc. In the stable branch  I couldn't handle http error (status code 401 etc) through a custom link. It may all solved now with the changes in beta branch. But please confirm your changes make it possible to catch errors at network level. thanks again. bsr

mainawycliffe:
Now that #445 has been merged, we can close this issue.

PR #445, opened 2019-10-26
feat(client): add error link

### LINKED ISSUE #419: Global Catch/Parse Error
Sometimes it is very useful to be able to catch specific error globally. For example, when you are using JWT auth and if JWT has expired or invalidated, user should be redirected to the login page. In current version of the package, we have to specify error handling method for each Query.

https://www.apollographql.com/docs/react/features/error-handling/#network-errors I think implementation of something similar to apollo would suffice.

jayjun:
Fixes #419 and #440.

Add `ErrorLink` which allows side effects such logging or signing out when a GraphQL or network error happens.

```dart
final ErrorLink errorLink = ErrorLink(errorHandler: (response) {
  if (response.graphQLErrors != null) {
    for (var error in response.graphQLErrors) {
      print('[GraphQL error]: Message: ${error['message']}, Location: ${error['locations']}, Path: ${error['path']}');
    }
  }

  if (response.networkError != null) {
    print('[Network error]: ${response.networkError}');
  }
});
```

Inspired by [apollo-link-error](https://github.com/apollographql/apollo-link/tree/master/packages/apollo-link-error).

#### Fixes / Enhancements

- Added `ErrorLink`.

#### Docs

- Document `ErrorLink` in README.md.

### DIFF
--- a/packages/graphql/lib/src/link/error/link_error.dart
+++ b/packages/graphql/lib/src/link/error/link_error.dart
@@ -1,0 +1,70 @@
+import 'dart:async';
+
+import 'package:graphql/src/link/link.dart';
+import 'package:graphql/src/link/operation.dart';
+import 'package:graphql/src/link/fetch_result.dart';
+import 'package:graphql/src/exceptions/exceptions.dart';
+import 'package:graphql/src/exceptions/graphql_error.dart';
+import 'package:graphql/src/exceptions/operation_exception.dart';
+
+typedef ErrorHandler = void Function(ErrorResponse);
+
+class ErrorResponse {
+  ErrorResponse({
+    this.operation,
+    this.fetchResult,
+    this.exception,
+  });
+
+  Operation operation;
+  FetchResult fetchResult;
+  OperationException exception;
+}
+
+class ErrorLink extends Link {
+  ErrorLink({
+    this.errorHandler,
+  }) : super(
+          request: (Operation operation, [NextLink forward]) {
+            StreamController<FetchResult> controller;
+
+            Future<void> onListen() async {
+              Stream stream = forward(operation).map((FetchResult fetchResult) {
+                if (fetchResult.errors != null) {
+                  List<GraphQLError> errors = fetchResult.errors
+                      .map((json) => GraphQLError.fromJSON(json))
+                      .toList();
+
+                  ErrorResponse response = ErrorResponse(
+                    operation: operation,
+                    fetchResult: fetchResult,
+                    exception: OperationException(graphqlErrors: errors),
+                  );
+
+                  errorHandler(response);
+                }
+                return fetchResult;
+              }).handleError((error) {
+                ErrorResponse response = ErrorResponse(
+                  operation: operation,
+                  exception: OperationException(
+                    clientException: translateFailure(error),
+                  ),
+                );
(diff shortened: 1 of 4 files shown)


packages/graphql/lib/src/link/error/link_error.dart
```diff
@@ -0,0 +1,63 @@
+import 'dart:async';
+
+import 'package:graphql/src/link/link.dart';
+import 'package:graphql/src/link/operation.dart';
+import 'package:graphql/src/link/fetch_result.dart';
+
+typedef ErrorHandler = void Function(ErrorResponse);
+
+class ErrorResponse {
```
micimize: Might be better for this to use an [`OperationException`](https://github.com/zino-app/graphql-flutter/blob/602aa7fb01a8a433638c4ec3b8140c931bc4bb6a/packages/graphql/lib/src/exceptions/operation_exception.dart) for consistency

jayjun: Sure, I like the idea. Having said that, `ErrorResponse` also contains the relevant `Operation` and `FetchResult` (and also a `NextLink` for [retries](https://github.com/apollographql/apollo-link/blob/26b10d052bfcbc036deb0ff46b17c8872560873d/packages/apollo-link-error/README.md#retrying-failed-requests) but in hindsight, should be left out since I didn’t implement that).

Should I,

1. replace `ErrorResponse` with `OperationException` and not pass the `Operation` and `FetchResult`? 

2. replace `ErrorResponse` with `OperationException` but pass the `Operation` and `FetchResult` as callback parameters instead?

3. keep `ErrorResponse` but nest `OperationException` as a field?

micimize: I'd nest the `OperationException`

jayjun: Done!

bsr203:
@HofmannZ @mainawycliffe I am not sure about release policy, but there are few exciting addition to this lib after last beta release. Anyway to get a  release with the recent changes. Thanks

mainawycliffe:
@bsr203 the release has been failing due to a pipeline error. I have a PR (#462) to fix that issue. We are supposed to release to beta daily, but due to the linting error introduced by another PR that hasn't happened.

mainawycliffe:
We should probably move to v3 beta and hope to fix all bug in a few weeks before releasing to stable.

packages/graphql/lib/src/link/error/link_error.dart
bsr203: Hi @jayjun I wonder why `NextLink` is left out. As in the linked apollo doc, I too facing issue with stale auth token, and like to retry the operation once get new token. Also, in the apollo implementation https://github.com/apollographql/apollo-link/blob/master/packages/apollo-link-error/src/index.ts , they don't do any special handling but just set the `NextLink` to `ErrorResponse`

jayjun: @bsr203 I left it out for the simple reason that I don’t use retries. Sounds like a bad reason but I feel if I don’t use it, I can’t do justice to the port.

Since you need it and know what changes to make, you should send a pull request.

Issue #422, opened 2019-09-24
Flutter For Web - Make Sure `graphql_flutter` works with flutter for Web


guiguan:
Because of this error:

```
[ +990 ms] Skipping compiling doc|lib/main_web_entrypoint.dart with ddc because some of its
                    transitive libraries have sdk dependencies that not supported on this platform:
                    
                    path_provider|lib/path_provider.dart
                    platform|lib/src/interface/local_platform.dart
                    
                    https://github.com/dart-lang/build/blob/master/docs/faq.md#how-can-i-resolve-skipped-compiling-warnings
```

However, the [dart only package](https://github.com/zino-app/graphql-flutter/tree/master/packages/graphql) seems to work in Flutter for Web

mainawycliffe:
I will look into this over next week. I am going to repurpose this issue so that we can use it to make sure flutter for the web works as expected.

mainawycliffe:
Support for flutter for web turns out is a little bit more complex. We currently don't have a way of switching between native APIs such as IO and browsers API such local storage, this makes it a little bit tricky to build one app that can be deployed to both native and web. We currently have some web APIs for web sockets and I believe in-memory cache but those I will have to test and we can make improvements if any are needed. 

It's still early for flutter for web, and with time, full support by this package should be possible.

guiguan:
Yes, a little bit tricky. I have a prototype working using your GraphQL package on both web and native. I had to use conditional imports for different implementations, such as MultipartFile. Then it seems working fine, even for the file uploads.

jascodes:
@mainawycliffe @guiguan Additionally I am looking for encrypted offline support. So was looking to make https://pub.dev/packages/sembast custom cache in between. How much work you guys think it is, if were to be done properly?

mainawycliffe:
@jascodes could you please open a new issue for this, it's better not to hijack another issue whose purpose is different.

mainawycliffe:
The good news is that InMemory cache is fully supported by flutter for web, I will PR some changes to make this seamless, we just need to make WebSockets compatible and we are headed somewhere.

brianschardt:
So right now this package still cannot be used for flutter for web? When are you guys expecting to make it partially compatible? Thanks for your hard work!

mainawycliffe:
[...]

PR #446, opened 2019-10-26
Flutter for Web: Correct in-memory cache imports

mainawycliffe:
I have moved some caching logic for determining the application path to the io specific classes implementing cache. This allows us `graphql_flutter` to be compatible with flutter to some extent by allowing us to use path_provider to get the application path only when importing `dart:io` is involved and not `dart:html`. This should not affect any of the public APIs and should continue to work as they are with the added advantage that it now compiles for flutter for web.

This aims to close this issue #422


packages/graphql/pubspec.yaml
```diff
@@ -20,6 +20,7 @@ dependencies:
   rxdart: ^0.22.0
   websocket: ^0.0.5
   quiver: ">=2.0.0 <3.0.0"
+  path_provider: ^1.0.0
```
micimize: I thought depending on `path_provider` makes a package web-incompatible? I think we'll need to move the `dart:io`-specific caching stuff to a 3rd packages

mainawycliffe: From my own tests, it doesn't seem to be the case. As long as it is not being used by flutter for web project, it will work. It seems they have improved it a bit, and it follows import paths instead. This allows you to use conditional imports for both `dart:io` and `dart:html` on the same project.

micimize:
Given that #590 indicated that we actually currently have web support, this seems unneeded now, so am closing

Issue #354, opened 2019-07-02
Cache invalidation/deletion options


micimize:
If I'm not mistaken, cache entities currently live forever. This is not good.

The [related apollo discussion](https://github.com/apollographql/apollo-feature-requests/issues/4)  probably has some good ideas

eusdima:
@micimize do you think some sort of simple `garbage collector` would work? 
Invalidating `cache keys` based on:
- `expire date`(default 7 days or something like that).
- `persistent storage limit`(128MB by default) and `in-memory limit`(should be configurable for device ram size and scale based on that).
- not enough space left for a new `key`.

Also adding to `cache` should have some boundaries: 
- If the `value` is too big(greater than storage limit), then it shouldn’t be added.

 I think `persistent storage` should have a `polling interval` to fetch and save new key that were created / updated / deleted and not the entire tree.

I could do most of the implementation if we agree on a spec and also fully test it. Maybe we could make it a third package for cache(reuse it in other projects and do-couple it from graphql)

micimize:
One thing we could do is just make the cache's `Map` storage injectable and default it to [`LruMap`](https://api.flutter.dev/flutter/quiver.collection/LruMap-class.html).

I think ideally, caching and persistent storage are concerns we externalize/generalize so that pretty much anything with a `Map` interface can be used, and we just provide defaults.

For example, we make a separate package, `persistent_maps`, that handles all the storage details, and auto-saves/re-saturates based on configuration, and then depend on it as a default. That way we don't have to worry about the eccentricities of caching.

serendipity1004:
@micimize If a use login using an account and then logs in with another account, the previous login account's cache is still available in second login account. I think we need a function to just simply invalidate the entire cache as well.

serendipity1004:
https://github.com/zino-app/graphql-flutter/blob/519381cd1c57515fd26cedb33c68a68d563a6eac/packages/graphql/lib/src/cache/in_memory_io.dart#L68-L72

is there a way to have this available without running `Query` or `Mutation` ?

mainawycliffe:
@serendipity1004 We should probably expose the reset method, maybe at the client level, so you can invalidate the cache easily.

And on a side note maybe we should switch to using an interface instead of an abstract class. This way, we all methods we need must be overridden by the implementing class and can't be ignored. This is because some cache have not implemented the reset() method override and am guessing an interface would guarantee some security in that regard.

micimize:
[...]

PR #447, opened 2019-10-28
cache.reset() added

serendipity1004:

#### Fixes / Enhancements

- exposed reset function on normalizedInMemoryCache and added test for it.
- previously requested as cache.clear() but cache.reset() function was already there in the abstract class. Did not want to break something so just went ahead with cache.reset();


serendipity1004:
#354

mainawycliffe:
@serendipity1004 can you change your target to and rebase it beta instead of master, as specified in the contributing guidelines

PR #449, opened 2019-10-28
Allow sending of requests using a Link, outside of Widgets context

eliezedeck:
Allows people to manually use a Link to send a request and get the result(s).

#### Fixes / Enhancements

- At this time, it seems we can only execute queries in the context of Flutter widgets. There are usecases that requires more than this, where outside of Widgets contexts, we need to run a query. This change allows just that by giving access to the `link.request()` which requires `Operation`.

#### Docs

- Exposes the `Operation` class


mainawycliffe:
Are you looking for something like [this](https://github.com/zino-app/graphql-flutter/tree/master/packages/graphql)?

eliezedeck:
I think I do ... I was trying to use the `Link` directly.

Issue #461, opened 2019-11-05
Fix CI/CD Pipeline


mainawycliffe:
We need to fix the CI/CD pipeline so that the linting test can pass.

mainawycliffe:
@micimize from what I can tell, it seems we have also not had a release since the end of August, can you confirm this?

mainawycliffe:
It seems the linting test is failing because of flutter analyze detecting use of deprecated `document` field of `QueryOptions`. This has been replaced by `DocumentNode` by the new AST Implementation. For this test to pass, we need to replace all instances of `document` (49 of them) with `DocumentNode`. In some cases like examples and tests, it's straight forward but since the functionality is deprecated and not removed, in instances it's not even possible to remove that.

mainawycliffe:
> @micimize from what I can tell, it seems we have also not had a release (beta) since the end of August, can you confirm this?

I have discovered the reason for this is due to failing tests. I hope PR #462 fixes the issue.

mainawycliffe:
@HofmannZ Sync beta with master is still failing due to a dependency resolution issue while release to beta is failing with the following message:
![Annotation 2019-11-12 122720](https://user-images.githubusercontent.com/12270550/68659179-d6a25080-0547-11ea-9342-2d391de1bb44.png)

PR #462, opened 2019-11-05
Attempt to fix Linting Errors Caused by Deprecating Document

mainawycliffe:
This is an attempt to fix linting errors caused by deprecating of Document field of the QueryOptions as highlighted by issue no #461 

All tests are passing now.

@micimize can you have a look at this when you can and see if you have any reservations. I had to add an [equitable](https://pub.dev/packages/equatable) package to make testing of Link class easier:

```dart
final linksFrom = Link.from([link1, link2, link3]);
final linksConcat = link1..concat(link2)..concat(link3);
expect(linksFrom, linksConcat);
```

### DIFF
--- a/examples/starwars/ios/Flutter/flutter_export_environment.sh
+++ b/examples/starwars/ios/Flutter/flutter_export_environment.sh
@@ -1,8 +1,0 @@
-#!/bin/sh
-# This is a generated file; do not edit or check into version control.
-export "FLUTTER_ROOT=C:\tools\flutter"
-export "FLUTTER_APPLICATION_PATH=C:\Users\mw\projects\opensource\graphql-flutter\examples\starwars"
-export "FLUTTER_TARGET=lib\main.dart"
-export "FLUTTER_BUILD_DIR=build"
-export "SYMROOT=${SOURCE_ROOT}/../build\ios"
-export "FLUTTER_FRAMEWORK_DIR=C:\tools\flutter\bin\cache\artifacts\engine\ios"
--- a/examples/starwars/ios/Flutter/flutter_export_environment.sh
+++ b/examples/starwars/ios/Flutter/flutter_export_environment.sh
@@ -1,11 +1,8 @@
 #!/bin/sh
 # This is a generated file; do not edit or check into version control.
-export "FLUTTER_ROOT=/Users/mjr/flutter"
-export "FLUTTER_APPLICATION_PATH=/Users/mjr/Documents/code/libraries/graphql-flutter/examples/starwars"
-export "FLUTTER_TARGET=/Users/mjr/Documents/code/libraries/graphql-flutter/examples/starwars/lib/main.dart"
+export "FLUTTER_ROOT=C:\tools\flutter"
+export "FLUTTER_APPLICATION_PATH=C:\Users\mw\projects\opensource\graphql-flutter\examples\starwars"
+export "FLUTTER_TARGET=lib\main.dart"
 export "FLUTTER_BUILD_DIR=build"
-export "SYMROOT=${SOURCE_ROOT}/../build/ios"
-export "FLUTTER_FRAMEWORK_DIR=/Users/mjr/flutter/bin/cache/artifacts/engine/ios"
-export "FLUTTER_BUILD_NAME=1.0.0"
-export "FLUTTER_BUILD_NUMBER=1"
-export "TRACK_WIDGET_CREATION=true"
+export "SYMROOT=${SOURCE_ROOT}/../build\ios"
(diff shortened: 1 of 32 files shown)


examples/starwars/ios/Flutter/flutter_export_environment.sh
```diff
@@ -1,11 +1,8 @@
...
+export "FLUTTER_APPLICATION_PATH=C:\Users\mw\projects\opensource\graphql-flutter\examples\starwars"
+export "FLUTTER_TARGET=lib\main.dart"
 export "FLUTTER_BUILD_DIR=build"
-export "SYMROOT=${SOURCE_ROOT}/../build/ios"
-export "FLUTTER_FRAMEWORK_DIR=/Users/mjr/flutter/bin/cache/artifacts/engine/ios"
-export "FLUTTER_BUILD_NAME=1.0.0"
-export "FLUTTER_BUILD_NUMBER=1"
-export "TRACK_WIDGET_CREATION=true"
+export "SYMROOT=${SOURCE_ROOT}/../build\ios"
+export "FLUTTER_FRAMEWORK_DIR=C:\tools\flutter\bin\cache\artifacts\engine\ios"
```
HofmannZ: Nit: we should ignore this file

mainawycliffe: Yes, let me see if I can update gitignore

examples/starwars/lib/reviews/review_page_list.dart
```diff
@@ -2,6 +2,7 @@ import 'package:flutter/material.dart';
 import 'package:graphql_flutter/graphql_flutter.dart';
 import 'package:starwars_app/reviews/review_subscription.dart'
     show DisplayReviews;
+import 'package:gql/language.dart';
```
micimize: I know apollo uses an analogous api with `graphql-tag`, but I find this api pretty non-ergonomic. Can we just re-export `parseString` as `gql` from `graphql/client` and `graphql_flutter`?

packages/graphql/analysis_options.yaml
```diff
@@ -0,0 +1,6 @@
+include: package:pedantic/analysis_options.yaml
+
+analyzer:
+  exclude:
+    - test/_data/**
+    - lib/**/**
```
micimize: is this to temporarily mute some error output? Looks pretty awkward 😅

mainawycliffe: Yea, it is temporary. As long as document is marked for deprecation then the linting test will not pass.

examples/starwars/lib/reviews/review_page_list.dart
mainawycliffe: so instead of parseString(query) we have something like gql(query)?

HofmannZ: @micimize I think, that’s a great good idea!

packages/graphql/analysis_options.yaml
HofmannZ: Can’t we override a specific listing rule?

packages/graphql/lib/src/link/link.dart
```diff
@@ -26,7 +27,7 @@ Link _concat(
   });
 }
 
-class Link {
+class Link extends Equatable {
```
micimize: >  hint • This class (or a class which this class inherits from) is marked as '@immutable', but one or more of its instance fields are not final: Link.request at lib/src/link/link.dart:30:7 • must_be_immutable

examples/starwars/lib/reviews/review_page_list.dart
micimize: right - the analogous idiom in apollo is the ``` gql`query {}` ``` template tag

packages/graphql/lib/src/link/link.dart
mainawycliffe: Let me work on that, we still need a way of testing that `link.from([link1, link2, link3])` is equal to `link1..concat(link2)..concat(link3)`. Maybe override equal behavior as Equatable does.

examples/starwars/lib/reviews/review_page_list.dart
mainawycliffe: Let me see what I can do.

mainawycliffe: it would have been amazing if dart supported [JavaScripts Tagged Template](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) Literals and we would have done it like apollo does `gql"query {}"` for now I guess it will have a function.

PR #469, opened 2019-11-12
Bump gql version, maybe fix nightly betas

micimize:
Might fix the nightly releases part of #461

Tried running circle ci locally, but there was a timeout during checkout. @HofmannZ do I need to set `GH_TOKEN` in the environment or something?


mainawycliffe:
@micimize it been a long time since we last had a release, maybe we should do a doc detailing changes that have been incorporated over the last 2 and a half months.

PR #474, opened 2019-11-15
WIP: fix AppLifecycleState.{suspending -> detached}

micimize:
Fix `graphql_flutter` on the most recent flutter version by updating app lifecycle

We need to restrict flutter sdk version with a lower bound, and mark this as a breaking change before merging

### DIFF
--- a/packages/graphql_flutter/lib/src/widgets/cache_provider.dart
+++ b/packages/graphql_flutter/lib/src/widgets/cache_provider.dart
@@ -61,7 +61,7 @@ class _CacheProviderState extends State<CacheProvider>
         client.cache?.save();
         break;
 
-      case AppLifecycleState.suspending:
+      case AppLifecycleState.detached:
         break;
 
       case AppLifecycleState.resumed:


micimize:
@mainawycliffe it happens to not be that big of a deal this time, but typically it's not a good idea to merge PR's that say "WIP" (Work In Progress)

mainawycliffe:
I am Sorry about that, didn't actually know that WIP was an acronym and since I was familiar with the previous PR, I thought it was done.

mainawycliffe:
I was actually thinking that since we are not using both suspending or detached states specifically, we are just ignoring them, we should use `default` instead. This way we maintain backward compatibility for those using different versions of flutter that support different `AppLifecycleState` states.

```dart
switch (state) {
  case AppLifecycleState.inactive:
    client.cache?.save();
    break;

  case AppLifecycleState.paused:
    client.cache?.save();
    break;

  case AppLifecycleState.resumed:
    client.cache?.restore();
    break;

  default:
    break;
}
```

PR #475, opened 2019-11-17
fix: use case else instead of detached or suspended

mainawycliffe:
Since we are not using `AppLifecycleState.detached` or `AppLifecycleState.suspended` for anything specific other than just breaking, I propose we a `case` `default` to handle the scenarios. This way, we will maintain compatibility with older and newer versions of flutter.

### DIFF
--- a/packages/graphql_flutter/lib/src/widgets/cache_provider.dart
+++ b/packages/graphql_flutter/lib/src/widgets/cache_provider.dart
@@ -61,11 +61,11 @@ class _CacheProviderState extends State<CacheProvider>
         client.cache?.save();
         break;
 
-      case AppLifecycleState.detached:
-        break;
-
       case AppLifecycleState.resumed:
         client.cache?.restore();
+        break;
+
+      default:
         break;
     }
   }


micimize:
I'm not sure this is a good idea. Right now, if a new `AppLifecycleState` is added that we should respond to, we'll see that our switch statement no longer covers all cases. If we use `default`, then we won't see it anymore.

Really, I think it would

mainawycliffe:
I understand that, but now we have a problem. `AppLifecycleState.Detached` is currently available in master only, which means, currently with the change, we have broken it for both stable and beta flutter, whom I think we should strive to make it always works for. With time, the change will be at Flutter beta, which means, again we can only support one version of flutter, until rolls to everyone else, which could take a few months, not sure though, as these are major changes to the flutter API. I get that, flutter analyze will give a warning if a new state added, but in such a scenario, if it was really important to our library am quite sure someone will catch it and file an issue. We can mitigate such a scenario by filing a new issue, adding a todo to revert to remove the default case, a few months down the line, when the change has hit the stable version and people have had time to upgrade. I think we should strive to make the package work with the widest range possible of flutter versions, and at the very least revert #474.

micimize:
@mainawycliffe you're right - I made a separate issue for my larger concerns in #477, but for now we need to not break the release and this is the way to do it

gunnardag:
When will this be available on pub.dev?

HofmannZ:
@gunnardag Should have been released automatically, but hang on Circle Ci, we fixed that now.

gunnardag:
@HofmannZ awesome, thanks

PR #481, opened 2019-12-02
fix: subscriptions reconnect

### LINKED ISSUE #407: GraphQLClient subscription will lost when re-connected to server
**What Happened**
In some reasons, I required to write my script through GraphQLClient and develop using Query, Mutation, Subscription. When I developing the subscription I found that when every connection closed and after re-connect, it didn't pass any query operation(StartOperation) to the server. Therefore, everything will no longer be received from the server.

**Expected behavior**
On re-connected will also able to receive the message from the server.

**Version**
- graphql v2.0.1

**Additional context**
I also checked the source code and found the reason happened in this line.
https://github.com/zino-app/graphql-flutter/blob/master/packages/graphql/lib/src/socket_client.dart#L254
If this line has gone, everything's works fine.

EugenePisotsky:
This PR resolves #407

#### Fixes / Enhancements

- updated `socket_client.dart` to keep map of `onListen` callbacks that are called again after reestablishing connection

p.s. I tried to reproduce it in test, but no luck (everything worked as expected). But manually it didn't work before this fix.

### DIFF
--- a/packages/graphql/lib/src/socket_client.dart
+++ b/packages/graphql/lib/src/socket_client.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:collection';
 import 'dart:convert';
 import 'dart:typed_data';
 import 'package:meta/meta.dart';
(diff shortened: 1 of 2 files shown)


mainawycliffe:
@kulture-rob thanks for the recommendation, I want to wrap up this so it can be included in the next version, so I have pushed the change myself.

mainawycliffe:
@EugenePisotsky can you please sign your commit.

EugenePisotsky:
@mainawycliffe maybe you can just pick my changes into another PR, because I'm really limited in time atm to prepare everything, sorry

mainawycliffe:
@EugenePisotsky no need to create a PR, we can go ahead and merge this, without the commit being signed.

Issue #363, opened 2019-07-18
Authorization Header "accessToken" was called on null


dali546:
**Describe the bug**
I am trying to create authentication for my app. If there are no user details present in the app (first time use) there is no token. I'm not too sure how to initialise my `AuthLink` when no token is present, but then change it once token is present (after login)

**To Reproduce**
1. Setup Client like so:
```dart
graphQLClient = ValueNotifier(
    GraphQLClient(
        cache: InMemoryCache(),
        link: AuthLink(getToken: () async {
          if (store.state.authUser.accessToken == null) {
            return "";
          } else {
            return "Bearer ${store.state.authUser.accessToken}";
          }
        }).concat(HttpLink(uri: "${store.state.serverURL}/graphql"))),
  );
```

2. Query for login:
```
static const login = r"""
    mutation Login($input: LoginInput!){
      login(
        data: $input
      )
      {
        access_token
        refresh_token
        expires_in
        token_type
        user {
          username
          email
          profile {
            forename
            surname
            birthday
            gender
          }
        }   
      }
    }
  """;
```
3. Login Mutation
```dart
QueryResult result = await graphQLClient.value.mutate(MutationOptions(
      document: Queries.login,
      variables: {
        "input": {
          "username": value['username'],
          "password": value['password'],
        },
      },
    ));
```
4. Error is found in `link_auth.dart`
`NoSuchMethodError: The getter 'accessToken' was called on null.
Receiver: null
Tried calling: accessToken`
```
#0      Object.noSuchMethod (dart:core-patch/object_patch.dart:50:5)
#1      initialiseGlobals.<anonymous closure> (package:community/vendor/globals.dart:21:36)
<asynchronous suspension>
#2      new AuthLink.<anonymous closure>.onListen (package:graphql/src/link/auth/link_auth.dart:18:52)
<asynchronous suspension>
#3      _runGuarded (dart:async/stream_controller.dart:805:24)
#4      _StreamController._subscribe.<anonymous closure> (dart:async/stream_controller.dart:684:7)
#5      _BufferingStreamSubscription._guardCallback (dart:async/stream_impl.dart:414:13)
#6      _StreamController._subscribe (dart:async/stream_controller.dart:683:18)
#7      _ControllerStream._createSubscription (dart:async/stream_controller.dart:818:19)
#8      _StreamImpl.listen (dart:async/stream_impl.dart:472:9)
#9      Stream.first (dart:async/stream.dart:1188:25)
[...]

PR #483, opened 2019-12-05
Support null value on getToken for authlink

Mosaal:
#### Fixes / Enhancements

The `getToken` method on the `AuthLink` class expects you to return a valid token as a string, yet it does not take into account that this token may not yet be defined whenever the method is executed. As such, it will throw a `[Malformed Authorization header: Undefined location]` when the invalid token is added as an `Authorization` header to the operation's context.

The only necessary change seems to be the following. From:
```
operation.setContext(<String, Map<String, String>>{
  'headers': <String, String>{'Authorization': token}
});
```
To:
```
if (token != null) {
  operation.setContext(<String, Map<String, String>>{
    'headers': <String, String>{'Authorization': token}
  });
}
```

### DIFF
--- a/packages/graphql/lib/src/link/auth/link_auth.dart
+++ b/packages/graphql/lib/src/link/auth/link_auth.dart
@@ -17,9 +17,11 @@ class AuthLink extends Link {
               try {
                 final String token = await getToken();
 
-                operation.setContext(<String, Map<String, String>>{
-                  'headers': <String, String>{'Authorization': token}
-                });
+                if (token != null) {
+                  operation.setContext(<String, Map<String, String>>{
+                    'headers': <String, String>{'Authorization': token}
+                  });
+                }
               } catch (error) {
                 controller.addError(error);
               }


mainawycliffe:
Doesn't accepting null defeat the purpose of an `AuthLink`? It is supposed to authenticate with a token and if the token is null, then you probably don't need an `AuthLink`.

franciscofsales:
I find this very useful. This would allow us to only create one GraphQL client that would support both the unlogged and logged in situation in an app. Currently I'm having to add complexity to create a different client for each scenario.

@mainawycliffe is currently there an alternative to handling the scenario where you might not yet be logged in and still want to use the client and then, after logging in, use an `AuthLink`?

mainawycliffe:
If you are using `graphql-flutter`, you can watch the auth status and once it changes just attach the `AuthLink` to your `GraphQLClient`, this will cause all widgets under `GraphQLProvider` to rebuild with the new client.

franciscofsales:
> If you are using `graphql-flutter`, you can watch the auth status and once it changes just attach the `AuthLink` to your `GraphQLClient`, this will cause all widgets under `GraphQLProvider` to rebuild with the new client.

I'm not, I'm doing an architecture in the likes of bloc... that's why I faced that challenge and actually there's more people facing the same issues like #363

mainawycliffe:
Even when using Bloc, the situation is the same. When state changes, let's say from Unauthenticated to authenticated, you can create a new Client with `AuthLink` attached to it. Then return the new client as part of the new state and rebuild the UI.

Issue #470, opened 2019-11-14
Migration guide?


serendipity1004:
A lot of changes have been made in beta3 it seems. I have upgraded and `result.error` has changed to `result.exception`, `update`s are not found in mutations, `onCompleted` is also not found. A migration guide to try out beta3 would be helpful. Or will the guide be available only for stable releases?

mainawycliffe:
There are so many changes, I have lost count, I will write a migration guide over the weekend.

On top of that, we need to jump to version 3, to allow people not ready for the upgrade to stay safely in version 2 since flutter follows semantic versioning. So, I think this we should be having a version 3.0.0.beta.1 instead. @micimize any ideas of how to bump the major version up?

micimize:
`update` and `onCompleted` moved to `MutationOptions`. The [changelog](https://github.com/zino-app/graphql-flutter/releases) will be useful for a migration guide

As for a new release - @HofmannZ correct me if I'm wrong, but I think we need to add the `BREAKING CHANGE:` footer to a commit,  which we've been neglecting 😬

HofmannZ:
@micimize yes, unfortunately you’re correct on that one.

mainawycliffe:
@HofmannZ can we force it?

HofmannZ:
It seems a bit weird to do a bump from beta 2.0.0 to beta 3.0.0 without releasing 2.0.0.

Looking at the release notes, promoting the current version to stable would already do a major version bump because `v2.1.1-beta.3` includes breaking changes.

mainawycliffe:
I don't understand, we already released versions 2.0.0 and 2.1.0, so since we are introducing breaking changes, we have to bump major versions.

v2.1.0 - https://github.com/zino-app/graphql-flutter/compare/v2.0.1...v2.1.0
v2.0.0 - https://github.com/zino-app/graphql-flutter/compare/v1.1.0...v2.0.0

HofmannZ:
Sorry for the comment earlier, we of course already have version 2.

Just reading through the [semantic release docs](https://semantic-release.gitbook.io/semantic-release/v/beta/usage/workflow-configuration#pre-release-branches) again. It looks like we should have merged features and fixes into master. And only push breaking changes to the beta branch.

mainawycliffe:
I think from v3, we should follow those guidelines. Bug fixes and patches go to master branch, while major features go to beta, then we can merge master into beta automatically daily just like we do.

HofmannZ:
[...]

PR #503, opened 2019-12-20
docs: update contributing guidelines

mainawycliffe:
This an update to contributing guidelines to direct that patches and other small fixes be deployed to the master branch, while new features especially major features should be deployed to beta. This will allow us to deploy fixes as soon as they are approved and not make people wait for a major release for bug fixes. As is already happening today, the CI will keep both master and beta in sync, so any changes to master will be automatically merged to PR. 

This addresses some of the issues raised in this PR - #470

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,7 +125,7 @@ from the main (upstream) repository:
   git push origin --delete my-fix-branch
   ```
 
-- Check out the beta branch:
+- Check out the master branch:
 
   ```shell
   git checkout master -f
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -120,7 +128,7 @@ from the main (upstream) repository:
 - Check out the beta branch:
```
micimize: this is inconsistent now - also idk if we need a section for after the PR, but we might as well make it a script they could just execute

mainawycliffe: Yeah, the section might be unnecessary, we copied this I think from Angular contributing guidelines. Could do with a bit more polishing.

mainawycliffe:
@HofmannZ can you approve this?

PR #510, opened 2019-12-26
pass NextLink to retry request

bsr203:
Fixes https://github.com/zino-app/graphql-flutter/issues/494


bsr203:
My attempt to implement retry feature using `NextLink` didn't succeed. I will update the original issue with the details and send a PR with example once get it working. Thanks

PR #517, opened 2019-12-31
Fix calling host getter on address which can be null

kolja-esders:
Fixes calling the `host` getter on an `address` that is `null`. This can occur when handling an exception because the client has no connectivity:

Screenshot 2019-12-31 at 01 05 01


mainawycliffe:
Rebase your request against beta and change your target branch to the same, please. And also remember to sign your commits, you can learn more [here](https://help.github.com/en/github/authenticating-to-github/about-commit-signature-verification)

mainawycliffe:
I am closing this in favor of #528.

PR #528, opened 2020-01-07
fix(client): fix(client): translateNetworkFailure fails with NPE when no route to host

laurisvan:
This one-liner change prevents a null pointer exception when host name resolution has failed. This can happen, e.g. when the phone has been left into flight mode, or when the network access has failed.

This bug was particularly nasty, as it prevented the normal error handling mechanism to operate (e.g. no OperationException would not be created).

### Breaking changes

- No breaking changes

#### Fixes / Enhancements

- Fixed translateNetworkFailure() method to not throw an error because of accessing null address at "failure.address.host"

#### Docs

- None added/updated


mainawycliffe:
Can you rebase your branch against the latest beta and also make sure you sign your requests.

PR #533, opened 2020-01-10
fix(flutter): widgets make unnecessary requests when dependencies change

fredericp:
Similar to #352 but it looks like it's stalled. I think this PR is more complete.

Widgets can get its `didChangeDependencies` method called multiple times even when no real changes have happened to its own dependencies. As a result, our query and mutation widgets may perform many unnecessary network fetches. In my scenario I get hundreds of unnecessary requests.

See [reddit discussion](https://www.reddit.com/r/FlutterDev/comments/c1911m/how_do_your_handle_fetching_data_inside/) and https://github.com/flutter/flutter/issues/39872 for more info.

The solution is to store the `client` in the widget state and check if it really changed before reinitializing the query.

### DIFF
--- a/packages/graphql_flutter/lib/src/widgets/mutation.dart
+++ b/packages/graphql_flutter/lib/src/widgets/mutation.dart
@@ -65,9 +65,6 @@ class MutationState extends State<Mutation> {
 
   // TODO is it possible to extract shared logic into mixin
   void _initQuery() {
-    client = GraphQLProvider.of(context).value;
-    assert(client != null);
-
     observableQuery?.close();
     observableQuery = client.watchQuery(_providedOptions.copy());
   }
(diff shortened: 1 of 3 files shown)


micimize:
@mainawycliffe this also looks good to me - if you've checked it out and tested it some we can merge it

nivekz:
May I request this PR be looked at again and hopefully be merged soon?

We have a production app that is severely affected by this bug (ie #575) and we have independently arrived at basically the same fix as this PR. So we are glad to find that this PR exists (otherwise we would be creating one 😄). I guess potentially many others are on the same boat. To have this PR merged and released soon can save us from all forking for a fix.

joseignaciopergolesi:
Sorry, could you show an example of how to make a query and not to do it three times?
I still can't do it with version 3.1.0-beta.2

joseignaciopergolesi:
It doesn't work for me using the latest beta version of the package. Could you upload an example of how to make a query using the wiget Query and not make three requests to the server?
Thank you

micimize:
@joseignaciopergolesi If [the basic example](https://github.com/zino-app/graphql-flutter/blob/master/packages/graphql_flutter/example/lib/graphql_widget/main.dart) produces unnecessary requests, please open a new issue extrapolating

nivekz:
@micimize @mainawycliffe I have a follow-up issue (currently as a comment in the original issue: https://github.com/zino-app/graphql-flutter/issues/575#issuecomment-612827953) and a proposed PR #601, and wonder if you could please take a look?

PR #534, opened 2020-01-11
allow initPayload to be a future

kulture-rob:
### Breaking changes

- none

#### Fixes / Enhancements

- Added the ability to use a Future for the initPayload of a SocketClient in the same way the AuthLink.getToken is. This allows a user to provide their bearer token more easily.

#### Docs

```dart
SocketClientConfig(
  initPayload: () async {
    FirebaseUser user = await FirebaseAuth.instance.currentUser();
    IdTokenResult idTokenResult = await user.getIdToken();

    return {
      "headers": {"Authorization": 'Bearer ${idToken.token}'},
    };
  },
)
```

### DIFF
--- a/packages/graphql/lib/src/socket_client.dart
+++ b/packages/graphql/lib/src/socket_client.dart
@@ -41,9 +41,12 @@ class SocketClientConfig {
 
   /// The initial payload that will be sent to the server upon connection.
   /// Can be null, but must be a valid json structure if provided.
-  final dynamic initPayload;
+  final FutureOr<dynamic> initPayload;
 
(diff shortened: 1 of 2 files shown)


kulture-rob:
@mainawycliffe sorry i missed the release of 3.0.0, circle ci was failing due to not being able to access my repo.

mainawycliffe:
@kulture-rob don't worry about, I think we should try and release very frequently, especially for nonbreaking changes. This is why I am proposing for non-breaking changes, they are merged to master directly. This way, they can be released weekly. The challenge with master branch is that it requires two reviewers, which can be quite difficult to get sometimes.

Issue #541, opened 2020-01-16
`Refetch` should not use cache


j-deng:
**Is your feature request related to a problem? Please describe.**
I have a long query list, want to use `cacheFirst` policy in case of refreshing every time when pushing or pop routes. But also want to use `refetch` in `RefreshIndicator`, for now it will use query options' fetch policy and will enable cache by default. refs: https://github.com/zino-app/graphql-flutter/blob/master/packages/graphql/lib/src/core/query_manager.dart#L222

**Describe the solution you'd like**
So I think we can copy and update options' fetch policy, it will not affect other functions.

j-deng:
Here is a possible fix https://github.com/zino-app/graphql-flutter/pull/542

Sh1d0w:
+1 about this feature. Currently `refetch` is useless if your fetch policy involves caching as it never retrieves the data from the server.

micimize:
Good point. I don't think there's any situation where you would want to `refetch` from the cache because [you've already got a stream of updates implicitly](https://github.com/apollographql/apollo-client/issues/5515#issuecomment-586382671)

prinze77:
+1 about this feature.

PR #542, opened 2020-01-16
Makes refetch to disable cache temporary

j-deng:
For #541 

### Breaking changes

- `refetch` use `cacheAndNetwork` to replace `cacheFirst`.
- Fix query lifecycle to pending on cache fetching.


packages/graphql/lib/src/core/query_manager.dart
```diff
@@ -219,7 +219,14 @@ class QueryManager {
   }
 
   Future<QueryResult> refetchQuery(String queryId) {
-    final WatchQueryOptions options = queries[queryId].options;
+    WatchQueryOptions options = queries[queryId].options;
+    if (options.fetchPolicy == FetchPolicy.cacheFirst) {
+      options = options.copy()
+        ..policies = Policies(
+          fetch: FetchPolicy.cacheAndNetwork,
```
Sh1d0w: Shouldn't it be FetchPolicy.networkOnly, as it always need to fetch the data from the server. How you changed it it will still return from cache?

j-deng: @Sh1d0w  Hey, what I found in [here](https://github.com/zino-app/graphql-flutter/blob/42d010c177f7fa752b9a0b5ca57548dd1cd3ad8f/packages/graphql/lib/src/core/query_options.dart#L33) that won't stop network fetching, and `cacheAndNetwork` won't affect the cache of origin fetch policy `cacheFirst`

packages/graphql/lib/src/core/observable_query.dart
```diff
@@ -123,11 +123,15 @@ class ObservableQuery {
...
-    lifecycle = _onDataSubscriptions.isNotEmpty
-        ? QueryLifecycle.SIDE_EFFECTS_PENDING
-        : QueryLifecycle.PENDING;
+    if (allResults.networkResult != null) {
+      // if onData callbacks have been registered,
+      // they are waited on by default
+      lifecycle = _onDataSubscriptions.isNotEmpty
+          ? QueryLifecycle.SIDE_EFFECTS_PENDING
+          : QueryLifecycle.PENDING;
+    } else if (allResults.eagerResult != null) {
```
j-deng: @Sh1d0w do you think it makes sense here, what it did is to satisfy [here](https://github.com/zino-app/graphql-flutter/blob/42d010c177f7fa752b9a0b5ca57548dd1cd3ad8f/packages/graphql/lib/src/core/observable_query.dart#L62) to make `cacheFirst` can be [refetch](https://github.com/zino-app/graphql-flutter/blob/42d010c177f7fa752b9a0b5ca57548dd1cd3ad8f/packages/graphql/lib/src/core/observable_query.dart#L81), I'm not very sure it won't have the side effect.

micimize: I don't think this change is necessary. Side effects are only possible with mutations, but refetch is only possible with queries, so there shouldn't be any kind of collision.

PR #557, opened 2020-02-07
fix "an operation already exists with this id"

kulture-rob:
#### Fixes

When triggering a new subscription while the websocket is reconnecting, GraphQL will throw an error of "an operation already exists with this id". If a subscription is adding during reconnect, the `onListen` is added to `_subscriptionInitializers` before the reconnect takes place. Then, when the connection is established, the original `onListen` triggers as well as the one that was stored off in `_subscriptionInitializers`. Since they are the same callback with the same id, an error is thrown.


packages/graphql_flutter/pubspec.yaml
```diff
@@ -8,7 +8,8 @@ authors:
 - Michael Joseph Rosenthal <rosenthalm93@gmail.com>
 homepage: https://github.com/zino-app/graphql-flutter/tree/master/packages/graphql_flutter
 dependencies:
-  graphql: ^3.0.1-beta.1
+  graphql:
+    path: ..\graphql
```
mainawycliffe: Do not commit the path for graphql, this is for development.

kulture-rob: Ahh whoops. I'll fix that

kulture-rob: Just pushed a fix

sokolej79:
Same issue.
It happens on Subscription widget when turn off/on internet.

Issue #549, opened 2020-01-24
Regression: flutter web no longer works


micimize:
Hi! Recently I've upgraded to the latest version of graphql.

I think the import of **"io_network_exception.dart"** on **"package:graphql/src/exceptions/exceptions.dart"** makes the package incompatible with web projects.

See the related [line](https://github.com/zino-app/graphql-flutter/blob/d71b553fe96231d14aa0dde905afca5e2d592866/packages/graphql/lib/src/exceptions/exceptions.dart#L2)

_Originally posted by @marcgraub in https://github.com/zino-app/graphql-flutter/issues/295#issuecomment-577743510_

micimize:
was a stupid mistake - should be an easy fix... at least for that line

binarypie:
This also broke dart web in general as it is we can't adopt the 3.x releases.

micimize:
closed by #590 and released in `3.1.0-beta.1`

Issue #522, opened 2020-01-06
Flutter web fails running because of cache property in GraphQLClient


chynamyerz:
**Describe the bug**
Failed to load resource: the server responded with a status of 404 (Not Found)
js_primitives.dart:32 ══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY ╞═══════════════════════════════════════════════════════════
js_primitives.dart:32 The following assertion was thrown building KeTaxiApp(dirty):
js_primitives.dart:32 Expected a value of type 'String', but got one of type '_Future<String>'
js_primitives.dart:32 
js_primitives.dart:32 Either the assertion indicates an error in the framework itself, or we should provide substantially
js_primitives.dart:32 more information in this error message to help you determine and fix the underlying cause.
js_primitives.dart:32 In either case, please report this assertion by filing a bug on GitHub:
js_primitives.dart:32   https://github.com/flutter/flutter/issues/new?template=BUG.md
js_primitives.dart:32 
js_primitives.dart:32 The relevant error-causing widget was:
js_primitives.dart:32   KeTaxiApp org-dartlang-app:///packages/ke_taxi/main.dart:14:10
js_primitives.dart:32 
js_primitives.dart:32 When the exception was thrown, this was the stack:
js_primitives.dart:32 package:dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 196:49      throw_
js_primitives.dart:32 package:dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 45:3        castError
js_primitives.dart:32 package:dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 426:10  cast
js_primitives.dart:32 package:dart-sdk/lib/_internal/js_dev_runtime/patch/core_patch.dart 661:12                _check
js_primitives.dart:32 package:graphql_flutter/src/caches.dart 15:8                                              new
js_primitives.dart:32 package:ke_taxi/main.dart 26:17                                                           build
js_primitives.dart:32 package:flutter/src/widgets/framework.dart 4364:28                                        build
js_primitives.dart:32 package:flutter/src/widgets/framework.dart 4296:15                                        performRebuild
js_primitives.dart:32 package:flutter/src/widgets/framework.dart 4020:5                                         rebuild
js_primitives.dart:32 package:flutter/src/widgets/framework.dart 4279:5                                         [_firstBuild]
[...]

PR #590, opened 2020-03-29
Starwars example web support

micimize:
Adds flutter web support to `examples/starwars`, thus confirming it's functionality.

Confirms that #422, #549, and #522 should be closed

Note that the server must have cors enabled for the flutter web port.

Will want to update the server once micimize/angel-starwars-test-server#3 is closed


micimize:
from discord convo:
> `graphql_flutter` does indeed work with flutter for web. The two gotchas we encountered were that the server needs cors enabled (at least from the flutter web port) and `Platform.is*` is not supported atm

PR #601, opened 2020-04-13
fix(flutter): Query.didUpdateWidget and policy overrides

nivekz:
The bug was first reported in https://github.com/zino-app/graphql-flutter/issues/575#issuecomment-612827953, quoted as follows:

> We found another bug that causes queries to be run multiple times in certain cases despite the fix in #533. That happens when:
> 1. Query is rebuilt by the parent. In the context of Navigator push/pop, this happens when Query has a stateful ancestor since every state seems to be built again when push/pop. See https://github.com/flutter/flutter/issues/11655
> 2. QueryState.didUpdateWidget is called, and it tries to compare the old and new options to decide whether to rebuild ObservableQuery.
> 3. Here is the actual bug: WatchQueryOptions.areEqualTo() does not consider default policy overrides. So when a query does not have fetch or error policies set, the equality check will fail, causing ObservableQuery to be rebuilt.

#### Fixes / Enhancements

- Fixed query options check in QueryState.didUpdateWidget by taking default policies into consideration.

Note:
- Added some test cases.
- Looks like the Mutation widget does not have the same problem since its implementation stores the previous query options BEFORE defaults overrides for comparison purpose. Nonetheless, I think comparing query options AFTER default overrides would be slightly better.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -7,9 +7,15 @@ flags:
         - packages/graphql/
 comment:
   require_changes: true # Only post a comment if coverage changes.
+
 coverage:
   status:
     project:
       default:
         threshold: 1% # Allow coverage to drop by up to 1% in a PR before marking it failed
     patch: off
+
+# don't factor tests into coverage scores
+ignore:
+  - packages/graphql_flutter/test
+  - packages/graphql/test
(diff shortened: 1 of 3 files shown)


nivekz:
The codecov report of drop in coverage is probably a codecov bug? This PR adds many more test cases ...

micimize:
@chicrock & @pepomps, y'all 🎉'd the original comment so I'm guessing you had the same issue. Could you try version `3.1.0-beta.3` to see if it's fixed?

chicrock:
@micimize Sound great!!🤩
I couldn't reproduce the issue. It's fixed.

nampdn:
I could confirm this issue has been fixed, when will it live in stable release?

PR #613, opened 2020-04-24
update to include initial state

vasilich6107:
In case of resolving from cache the 'loading' initial state show the
loading indicator each time the query renders.

Moved cache instantiation out of `_client` to make cache work.
Otherwise cache instance was always recreated.

### Breaking changes

- no breaking changes

#### Fixes / Enhancements

- Updated - bloc sample to include initial state
- Fixed - moved cache instantiation out of `_client` to make cache work


vasilich6107:
@micimize @mainawycliffe 

Now it's your turn ;-)

vasilich6107:
Thanks!

PR #630, opened 2020-05-09
fix: in memory html fails if it gets a future

olafura:
#### Fixes / Enhancements
Currently the star wars example is broken when using flutter web, this fixes it.

I'm not able to get the star wars server running because the angle graphql server is trying to access the message of a typeerror. But I tested it with a javascript version. Don't know if it functions completely but at least it doesn't break as badly.


micimize:
outdated by `4.0.0-beta.1` – please try there and if there's still an issue we'll go from there

micimize:
This was for #717 I believe

PR #635, opened 2020-05-13
fix(client): FetchMoreOptions bug with operator precedence

nivekz:
This PR fixes a bug in FetchMoreOptions's constructor that will throw when documentNode parameter is non-null, which is due to `??` having higher precedence than `? :`.

### Breaking changes

- There should be NO breaking change because:
  - When documentNode is non-null, old code will throw (but wrongly), new code will not
  - When documentNode is null, both code will give the same result

Note: I feel that it might be better for this PR to merge into master instead of into beta, since it is a pure non-breaking bug fix.


micimize:
@nivekz would merge this into `master` if you want to handle changing the base – github couldn't do it automatically

nivekz:
@micimize Yes, I have changed the base to `master`.

PR #681, opened 2020-06-27
fix: run onError callback before onCompleted

ronaldcurtis:
Fixes https://github.com/zino-app/graphql-flutter/issues/680 https://github.com/zino-app/graphql-flutter/issues/643

**Fixes / Enhancements**
As a developer executing a mutation request
When I receive an error from the server
Then I should be able to react to this error in the onError callback, before the onCompleted callback is run.

**Example use case:**
In my app, I allow the user to update his/her username. And then I Navigator.of(context).pop to the previous route after the mutation. But I don't want to pop the screen if there's an error. Instead I want to be able to show some kind of error message.

### DIFF
--- a/packages/graphql/lib/src/core/query_options.dart
+++ b/packages/graphql/lib/src/core/query_options.dart
@@ -178,7 +178,7 @@ class MutationCallbacks {
   // callbacks will be called against each result in the stream,
   // which should then rebroadcast queries with the appropriate optimism
   Iterable<OnData> get callbacks =>
-      <OnData>[onCompleted, update, onError].where(notNull);
+      <OnData>[onError, update, onCompleted].where(notNull);
 
   // Todo: probably move this to its own class
   OnData get onCompleted {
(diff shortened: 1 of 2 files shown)


ronaldcurtis:
I've confirmed that this fixes the problem described

micimize:
It's not just that we should call `onCompleted` first – we should wrap `onCompleted` and `update` in exception handling, and call `onError` with any errors thrown, closer to what [apollo does](https://github.com/apollographql/apollo-client/blob/4a74e29cb64c7d013d55f05873d8567dd4b95aee/src/react/data/MutationData.ts#L71-L79) 
```typescript
    return this.mutate(mutationFunctionOptions)
      .then((response: FetchResult<TData>) => {
        this.onMutationCompleted(response, mutationId);
        return response;
      })
      .catch((error: ApolloError) => {
        this.onMutationError(error, mutationId);
        if (!this.getOptions().onError) throw error;
      });
```

PR #696, opened 2020-07-24
Beta synced with master

micimize:
Rebasing `beta` against `master` in prep for merge


micimize:
Expected there to be deeper-cutting changes, but ok. Will see how this changes the diff of `beta -> master`

PR #702, opened 2020-07-27
Fix mutation closing

### LINKED ISSUE #700: v4 Mutation widget doesn't work as expected
**Describe the bug**
In v4 if we have built mutation widget and we fired a mutation once => everything is okay. On completed callback fired, loading state changes.
However, if we didn't dispose of the mutation widget and we will try to fire the same mutation again => on completed callback is not fired, the loading state is not configured, however mutation itself is executed successfully.

I have simulated this in simple flutter create app. 

**To Reproduce**
1. Copy this snippet to your main.dart file
```

void main() {
  runApp(MyApp());
}

class MyApp extends StatelessWidget {
  // This widget is the root of your application.
  @override
    final HttpLink httpLink = HttpLink("https://graphqlzero.almansi.me/api");
    final Link link = httpLink;

    var cache = GraphQLCache();

    var graphQLClient = GraphQLClient(
      cache: cache,
      link: link,
    );

    var client = ValueNotifier(graphQLClient);

    return GraphQLProvider(
      client: client,
      child: CacheProvider(
        child: MaterialApp(
          title: 'Flutter Demo',
          theme: ThemeData(
            // This is the theme of your application.
            //
            // Try running your application with "flutter run". You'll see the
            // application has a blue toolbar. Then, without quitting the app, try
            // changing the primarySwatch below to Colors.green and then invoke
            // "hot reload" (press "r" in the console where you ran "flutter run",
            // or simply save your changes to "hot reload" in a Flutter IDE).
            // Notice that the counter didn't reset back to zero; the application
            // is not restarted.
            primarySwatch: Colors.blue,
            // This makes the visual density adapt to the platform that you run
            // the app on. For desktop platforms, the controls will be smaller and
            // closer together (more dense) than on mobile platforms.
            visualDensity: VisualDensity.adaptivePlatformDensity,
          ),
          home: MyHomePage(title: 'Flutter Demo Home Page'),
        ),
      ),
    );
  }
}

class MyHomePage extends StatefulWidget {
  MyHomePage({Key key, this.title}) : super(key: key);

  // This widget is the home page of your application. It is stateful, meaning
  // that it has a State object (defined below) that contains fields that affect
  // how it looks.

  // This class is the configuration for the state. It holds the values (in this
  // case the title) provided by the parent (in this case the App widget) and
  // used by the build method of the State. Fields in a Widget subclass are
  // always marked "final".

  final String title;

  @override
  _MyHomePageState createState() => _MyHomePageState();
}

  int _counter = 0;

  void _incrementCounter() {
    setState(() {
      // This call to setState tells the Flutter framework that something has
      // changed in this State, which causes it to rerun the build method below
      // so that the display can reflect the updated values. If we changed
      // _counter without calling setState(), then the build method would not be
      // called again, and so nothing would appear to happen.
      _counter++;
      print(_counter);
    });
  }

  @override
    // This method is rerun every time setState is called, for instance as done
    // by the _incrementCounter method above.
    //
    // The Flutter framework has been optimized to make rerunning build methods
    // fast, so that you can just rebuild anything that needs updating rather
    // than having to individually change instances of widgets.
    return Scaffold(
      appBar: AppBar(
        // Here we take the value from the MyHomePage object that was created by
        // the App.build method, and use it to set our appbar title.
        title: Text(widget.title),
      ),
      body: Center(
        // Center is a layout widget. It takes a single child and positions it
        // in the middle of the parent.
        child: Column(
          // Column is also a layout widget. It takes a list of children and
          // arranges them vertically. By default, it sizes itself to fit its
          // children horizontally, and tries to be as tall as its parent.
          //
          // Invoke "debug painting" (press "p" in the console, choose the
          // "Toggle Debug Paint" action from the Flutter Inspector in Android
          // Studio, or the "Toggle Debug Paint" command in Visual Studio Code)
          // to see the wireframe for each widget.
          //
          // Column has various properties to control how it sizes itself and
          // how it positions its children. Here we use mainAxisAlignment to
          // center the children vertically; the main axis here is the vertical
          // axis because Columns are vertical (the cross axis would be
          // horizontal).
          mainAxisAlignment: MainAxisAlignment.center,
          children: <Widget>[
            Text(
              'You have pushed the button this many times:',
            ),
            Text(
              '$_counter',
              style: Theme.of(context).textTheme.headline4,
            ),
          ],
        ),
      ),
      floatingActionButton: Mutation(
        options: MutationOptions(
          document: gql('''
mutation CreateAlbum(\$title: String!, \$userId: ID!) {
  createAlbum(input:{title: \$title, userId: \$userId}) {
    id
    title
  }
} '''),
          onCompleted: (dynamic resultData) async {
            if (resultData != null) {
              print('onCompleted fired');
            }
          },
        ),
        builder: (
          RunMutation runMutation,
          QueryResult result,
        ) =>
            FloatingActionButton(
          onPressed: () {
            runMutation({
              'title': 'Hi',
              'userId': '1',
            });
            _incrementCounter();
          },
          tooltip: 'Increment',
          child: Icon(Icons.add),
        ),
      ), // This trailing comma makes auto-formatting nicer for build methods.
    );
  }
}
```

2. Add
`  graphql_flutter: 4.0.0-alpha.2` to your `pubspec.yaml` file. 

3. Try pressing the plus button multiple times. You will see that counter executes successfully however `onCompleted` callback was not fired. (see image)
image

**Expected behavior**
All callbacks are fired. State of mutation result has changes. 

**Screenshots**

**Tablet:**
 - Android Lenovo M10
 - v9

micimize:
Closes #700 by not closing mutations after callbacks run

changes published as `4.0.0-alpha.3` – @Sanjikir lmk if it fixes things for you

#### Fixes / Enhancements

- don't close `ObservableQuery` after callbacks fire while side effects are pending
- update `graphql_flutter/example` to work with most recent flutter

### DIFF
--- a/packages/graphql/CHANGELOG.md
+++ b/packages/graphql/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 4.0.0-alpha.3 (2020-07-27)
+* **client**: don't close mutations after callbacks ([2ba6c74](https://github.com/zino-app/graphql-flutter/commit/2ba6c743a7317c3df05c3f1c5e8e3e3cd44d6827))
+* **examples**: update ios files for graphql_flutter/example ([5b6e3d0](https://github.com/zino-app/graphql-flutter/commit/5b6e3d06a7dc56888dcbfc4c395ea51985c10f1c))
+
+
 # 4.0.0-alpha.2 (2020-07-24)
 * **client**: simplified AuthLink ([0b3fbd9](https://github.com/zino-app/graphql-flutter/commit/0b3fbd9a4d3d0f9bded2bd9a9fdf26e3bfe983df))
 * **docs**: direct cache access in changelog ([1f21927](https://github.com/zino-app/graphql-flutter/commit/1f2192710531a920be6c7df1b22da105c8cdc11c))
--- a/packages/graphql/lib/src/core/observable_query.dart
+++ b/packages/graphql/lib/src/core/observable_query.dart
@@ -289,9 +289,9 @@ class ObservableQuery {
           lifecycle = QueryLifecycle.completed;
           close();
         }
+        // the mutation has been completed, but disposal has not been requested
         if (lifecycle == QueryLifecycle.sideEffectsPending) {
           lifecycle = QueryLifecycle.completed;
-          close();
         }
       }
     }
--- a/packages/graphql/pubspec.yaml
+++ b/packages/graphql/pubspec.yaml
@@ -2,7 +2,7 @@ name: graphql
 description:
   A stand-alone GraphQL client for Dart, bringing all the features from
   a modern GraphQL client to one easy to use package.
-version: 4.0.0-alpha.2
+version: 4.0.0-alpha.3
 homepage: https://github.com/zino-app/graphql-flutter/tree/master/packages/graphql
 dependencies:
   meta: ^1.1.6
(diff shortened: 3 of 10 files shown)


SanjiKir:
@micimize Works like a charm! Thanks man!

Issue #744, opened 2020-10-13
Throw exceptions on malformed / mismatched data


micimize:
Right now `writeQuery`/`readQuery` will return `null` if the data structure doesn't match the operation. we should throw some exceptions based on this.

Actually, we should probably run validation on every `write*` to the `cache` to ensure manual cache access is valid as well

Related: https://github.com/gql-dart/ferry/issues/102

micimize:
Closed by #754 and configurable by `partialDataPolicy`

Issue #718, opened 2020-09-01
Cached data is not available on the QueryResult when there is a network error.


jifalops:
**Describe the bug**
Cached data is not available on the QueryResult when there is a network error.

**To Reproduce**
1. Set-up
```dart
GraphQLProvider(
      client: ValueNotifier<GraphQLClient>(
        GraphQLClient(
          cache: OptimisticCache(dataIdFromObject: _idFromObject),
          link: myLink,
        ),
      ),
      child: CacheProvider(
        child: MaterialApp(
  ...
```
2. Run a query and get the results.
```dart
class MyQuery extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return Query(
      options: QueryOptions(
        // query is of a type generated by the `artemis` package
        documentNode: query.document,
      ),
      builder: (
        QueryResult result, {
        Future<QueryResult> Function() refetch,
        FetchMore fetchMore,
      }) {
        if (result.hasException) {
          return Text('${result.exception}');
        } else if (result.loading) {
          return Center(child: CircularProgressIndicator());
        }

        final items = query.parse(result.data);

        return Column(
          children: [
            ...items.map((item) => ListTile(
                  title: Text(item.name),
                  subtitle: Text('${item.typeName} ${item.id}'),
                ))
          ],
        );
      },
    );
  }
}
```
3. Enable airplane mode on the device
4. Press home and open another app (so the cache write is triggered)
5. Restart the app (hot restart). The result is now the network error and no cached data is available.

**Expected behavior**
Enabling airplane mode and hot-restarting the app should allow it to return the cached results.

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04

**Smartphone (please complete the following information):**
 - Device: Android emulator API 29

**Additional context**
Flutter beta 1.21.0-9.2.pre

micimize:
Can you try this with `4.0.0-alpha.5`?

micimize:
This is outdated / likely solved by the cache rework `4.0.0-beta.1` (#648). Also this may just be because of the hot restart.

Please make a reproduction that doesn't involve live reload for testing

micimize:
looked at this again and added a mechanism for carrying forward errors in #754

jifalops:
i actually solved it with an HTTP link that caches queries. I think that is what's right for my use case.

micimize:
Closed by #754

Issue #687, opened 2020-07-16
Flutter Web Throws Error "Expected a value of type 'String', but got one of type '_Future<String>'"


agdsat134:
**Describe the bug**
Getting the following error when compiling Star Wars example for Flutter Web.
```
══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY ╞═══════════════════════════════════════════════════════════
The following TypeErrorImpl was thrown building MyApp(dirty):
Expected a value of type 'String', but got one of type '_Future<String>'

  MyApp file:///F:/graphql-flutter/examples/starwars/lib/main.dart:21:23

When the exception was thrown, this was the stack:
C:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/eC:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 62:3        castError
C:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 442:10  cast
ypes.dart 412:9        as
packages/graphql/src/cache/normalized_in_memory.dart 21:8
new
packages/graphql/src/cache/optimistic.dart 66:8
new
packages/graphql_flutter/src/caches.dart 33:8
new
packages/starwars_app/client_provider.dart 15:31
get cache
C:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 816:6   get
packages/starwars_app/client_provider.dart 38:14
clientFor
packages/starwars_app/client_provider.dart 51:17
new
packages/starwars_app/main.dart 26:12
build
packages/flutter/src/widgets/framework.dart 4585:28
build
packages/flutter/src/widgets/framework.dart 4511:15
performRebuild
packages/flutter/src/widgets/framework.dart 4227:5
rebuild
packages/flutter/src/widgets/framework.dart 4490:5
[_firstBuild]
packages/flutter/src/widgets/framework.dart 4485:5
mount
packages/flutter/src/widgets/framework.dart 3455:13
inflateWidget
packages/flutter/src/widgets/framework.dart 3223:18
updateChild
packages/flutter/src/widgets/binding.dart 1132:16
[_rebuild]
[...]

Issue #683, opened 2020-07-02
graphql-flutter not working in flutter web


minhnhatvdl:
**Describe the bug**
I am using graphql-flutter v3.0.1 for a project of flutter web. Last week, it worked very well. Now, it no longer works.
<img width="392" alt="Capture d’écran 2020-07-02 à 20 51 24" src="https://user-images.githubusercontent.com/12047824/86398403-d3f69600-bca5-11ea-9586-3280d8bdc960.png">

I am using flutter in channel beta with version 1.12.13+hotfix.6

MostafaHassan:
I got the same error in beta, dev and master. 
Note: I only got the error on web server.
So to reproduce, run flutter build web

ductranit:
I had the same issue. Notice for more details: it looks like there is a problem with dependency connectivity 0.4.9 which uses connectivity_for_web 0.3.0 because I got 404 error for connectivity_for_web.dart

GP4cK:
Here is the root cause: https://github.com/flutter/flutter/issues/60855.
You can also add this to your `pubspec.yml`:

```yml
dependency_overrides:
  connectivity: ^0.4.9
  connectivity_for_web: ^0.3.1
```

micimize:
Closed by #754

PR #754, opened 2020-10-24
Structural data validation on cache writes and more

### LINKED ISSUE #747: [v4] Both result data and exception are sometimes null on network result
**Describe the bug**
Just upgraded to graphql_flutter: `^4.0.0-beta.3`. What I did  is a normal query as I did before.

I'm not sure what is the problem.
```graphql
query {
  userByToken {
    newOrders(query: {skip: 0, sortBy: "createdAt", order: 1}) {
      id
      dishMetas {
        dish {
          id
          name
          store {
            id
            name
            description
            avatar
            district {
              name
            }
          }
          tag
          price
          avatar
        }
        quantity
        remark
      }
      status
      initiator {
        id
        name
        tel
        address
      }
      deliverPrice
      createdAt
    }
  }
}
```
I do the same query in postman which is all fine. 
I wonder in what situation this data would be null? isLoading is false, and there is no exception. Having worked on it for hours and have no clue.

micimize:
Closes #747, #744, #718, #683, #687, #616, #405 

depends on https://github.com/gql-dart/ferry/pull/104

### `result.data` carry forward on exception (#718)
for `watchQuery`, when an exception results in `null` data, the previous result will be used by default. Can be disabled by `WatchQueryOptions(carryForwardDataOnException: false)`

### only rebroadcast when !deep equals by default (#616)
`alwaysRebroadcast` can be used to get the old behavior back.

### strict cache writes (semi-breaking)
cache writes are now strict and throw PartialDataException (from normalize) or a CacheMisconfigurationException if the structures seem like they should be valid (configurable by `partialDataPolicy`)

At link execution time, one of the following exceptions can be thrown:
* CacheMisconfigurationException if the structure seems like it should
   write properly.
* UnexpectedResponseStructureException if the server response looks malformed.
* MismatchedDataStructureException in the event of a malformed
   optimistic result.
* If write succeeds but `readQuery` then returns `null`, a `CacheMissException` will be added and **the data will not be overwritten**

before, one could attempt to write any data structure to the cache, and it would write whatever document-matching data there was.
Now, if one attempts to write structurally partial data, appropriate
exceptions will be thrown.

### fechMore updates
* Corresponding to the above, `FetchMoreOptions.partial` is added for safer update operations
* unrelatedly, `UpdateQuery` is now strictly typed with `Map`s

### Skip cache writes when errors && ErrorPolicy.none (#405)
```
  /// If we have no data, we skip caching, thus taking [ErrorPolicy.none]
  /// into account
```

I also verified `examples/starwars` works on web and bumped the connectivity plugin to silence an error.


micimize:
This depends on https://github.com/gql-dart/ferry/pull/104 getting merged, but can be tried out like so:
```yaml
dependency_overrides:
  graphql:
    git: 
      url: git@github.com:micimize/graphql-flutter.git
      ref: norm_data_handling
      path: packages/graphql
```

micimize:
Note to self: Update Exceptions section with strictness errors https://github.com/zino-app/graphql-flutter/tree/beta/packages/graphql#exceptions

EightRice:
I got here cause Uncaught (in promise) Error: Expected a value of type 'String', but got one of type 'NativeJavaScriptObject'
and this was supposed to be the solution... I have no idea what all this text is but if anyone can kindly tell me what to do about my error, I'd be really grateful.

micimize:
@EightRice open a new issue – that sounds like a web-related issue and at first blush doesn't look necessarily related to strict cache writes (maybe that's why the error was surfaced, but not the root cause)

PR #758, opened 2020-10-29
Update link_http.dart

lohitbura:
Storing response  headers in headerGl

### Breaking changes

- Broke ... because ... .

#### Fixes / Enhancements

- Fixed ... was ... .
- Added ... .
- Updated ... .

#### Docs

- Added ... .
- Updated ... .


micimize:
While v4 is in beta I'm only merging hotfixes into master, but also we're not going to just write headers to a random global variable, sorry

PR #772, opened 2020-11-17
Check for nullable keepAliveSubscription

phuchuynhStrong:
Check for nullable ```keepAliveSubscription``` when disposing ```SocketClient```.

**Fixes**
- Fixed ```dispose``` function of ```SocketClient```, it didn't check the nullable ```keepAliveSubscription``` before put it into ```Future.wait```.  If ```SocketClientConfig``` constructed with null ```inactivityTimeout```, ```keepAliveSubscription``` won't be initialized which results in the exception in ```Future.wait```.

### DIFF
--- a/packages/graphql/lib/src/links/websocket_link/websocket_client.dart
+++ b/packages/graphql/lib/src/links/websocket_link/websocket_client.dart
@@ -261,15 +261,12 @@ class SocketClient {
   Future<void> dispose() async {
     print('Disposing socket client..');
     _reconnectTimer?.cancel();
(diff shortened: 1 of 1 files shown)


micimize:
if this is an issue it would be better to apply it to all the items in the list, as they all appear to be nullable. Something like ` Future.wait([...].where((closing) => closing != null)).toList())`

phuchuynhStrong:
Yup. It's better approach to handle this func. I updated the code.

micimize (approved):

Issue #771, opened 2020-11-12
ObservableQuery lifecycle hangs in pending for non-networked fetch policies


micimize:
We set the lifecycle to pending without consideration of whether we're waiting for anything:
https://github.com/zino-app/graphql-flutter/blob/2efce9f15e59208d86a2c4b1c9e0794cc28ea680/packages/graphql/lib/src/core/observable_query.dart#L192-L194

This was found by @blaugold because it breaks `cacheFirst -> refetch()` behavior: https://github.com/zino-app/graphql-flutter/issues/346#issuecomment-725305734

micimize:
closed by #777

PR #777, opened 2020-11-30
fix(): fix ObservableQuery.lifecycle for cache only results

blaugold:
Fix for #771


micimize (approved):
LGTM – I _might_ have preferred to base the condition on the policy itself, but idk if that would be better

PR #778, opened 2020-12-04
Implement toJson encoding for DateTime objects.

LikeCarter:
DateTime does not have a toJson() default encoding.

Using this object in GQL mutations always fails.

### Breaking changes

- None.

#### Fixes / Enhancements

- Fixed DateTime objects being used in GQL queries/mutations.
- Added toJson() support for DateTime.
- Updated toKey() function.

### DIFF
--- a/packages/graphql/lib/src/core/raw_operation_data.dart
+++ b/packages/graphql/lib/src/core/raw_operation_data.dart
@@ -85,6 +85,12 @@ class RawOperationData {
       if (isIoFile(object)) {
         return object.path;
       }
+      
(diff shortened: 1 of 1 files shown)


micimize:
Development on `master` is frozen aside from hotfixes at the moment. `beta` doesn't have the same `toKey` mechanism anymore, and expected variables to  be json serializable already (although a [custom serializer](https://github.com/gql-dart/gql/blob/402d40573031b2206128a8d4e384f96fef46f55a/links/gql_http_link/lib/src/link.dart#L74) can be provided).

PR #781, opened 2020-12-10
Use Equatable/EquatableMixin in classes to help in mock testing

lesliearkorful:
This pull request seeks to add equality to classes passed to the client.
The reason is simply to allow the client to be directly mocked with expected results.
Because of the lack of equality, arguments passed to `client.mutate()`, ` client.query()` and ` client.watchQuery()` are failed to be matched by the `when` function in `mockito`.

### Breaking changes

- There are no breaking changes

#### Fixes / Enhancements

- `QueryOptions`, `MutationOptions`, `WatchQueryOptions` have been updated in `query_options.dart`. These classes now use the `EquatableMixin`
- `QueryResult` has been updated to extend  `Equatable`

#### Docs

- No documentation was needed and was not added

#### Test
- Test added in `tests/graphql_client_mock_test.dart`
An example below
```dart
    // this works now. It was previously failing
    test('Login Mutation', () async {
      final email = "test@user.com";
      final password = "password";
      final options = MutationOptions(
        documentNode: document,
        variables: {'email': email, 'password': password},
      );
      when(client.mutate(options))
          .thenAnswer((_) => Future.value(loginResponse));
      final res = await client.mutate(
        MutationOptions(
          documentNode: document,
          variables: {'email': email, 'password': password},
        ),
      );
      expect(res, loginResponse);
     // result is true and both are equal
    });
```

### DIFF
--- a/packages/graphql/lib/src/core/query_options.dart
+++ b/packages/graphql/lib/src/core/query_options.dart
@@ -1,3 +1,4 @@
+import 'package:equatable/equatable.dart';
 import 'package:gql/ast.dart';
 import 'package:gql/language.dart';
 import 'package:graphql/client.dart';
@@ -105,7 +106,7 @@ class BaseOptions extends RawOperationData {
 }
 
 /// Query options.
-class QueryOptions extends BaseOptions {
+class QueryOptions extends BaseOptions with EquatableMixin {
   QueryOptions({
     @Deprecated('The "document" option has been deprecated, use "documentNode" instead')
         String document,
@@ -129,6 +130,18 @@ class QueryOptions extends BaseOptions {
   /// The time interval (in milliseconds) on which this query should be
   /// re-fetched from the server.
   int pollInterval;
+
+  @override
+  List<Object> get props => [
+        documentNode?.definitions,
+        documentNode?.span?.sourceUrl,
+        variables,
+        fetchPolicy,
+        errorPolicy,
+        optimisticResult,
+        pollInterval,
+        context,
+      ];
 }
 
 typedef OnMutationCompleted = void Function(dynamic data);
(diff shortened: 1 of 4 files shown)


micimize:
So, I appreciate the time, effort, and spirit of this PR, but there are a few issues that mean I probably won't merge this:
1. `Equatable` is designed for use with immutable objects, and (unfortunately) our api is not yet immutable. This is a more general issue that I discovered while pursuing the potential use of Equatable before:
> [Note](https://github.com/felangel/equatable#usage): Equatable is designed to only work with immutable objects so all member variables must be final (This is not just a feature of Equatable - [overriding a hashCode with a mutable value can break hash-based collections](https://dart.dev/guides/language/effective-dart/design#avoid-defining-custom-equality-for-mutable-classes)).
2. Development work is done off of the `beta` branch, which is now fairly different compared to `master`. Here's a relevant note that may be of interest:
https://github.com/zino-app/graphql-flutter/blob/4877b39fda1c2317cb8d81fc537ee28e7b85bce6/packages/graphql/lib/src/core/_data_class.dart#L7-L11
3. For testing purposes, mockito has structural matchers like [`argThat(matcher)`](https://pub.dev/documentation/mockito/latest/mockito/argThat.html), which I _think_ are used like so, based on my experience [elsewhere](https://github.com/zino-app/graphql-flutter/blob/4877b39fda1c2317cb8d81fc537ee28e7b85bce6/packages/graphql/test/graphql_client_test.dart#L202-L207):
```dart
    test('Login Mutation', () async {
      final email = "test@user.com";
      final password = "password";
      final options = MutationOptions(
        documentNode: document,
        variables: {'email': email, 'password': password},
      );
      when(client.mutate(argThat(isA<MutationOptions>().having(
          (o) => o.variables['email'],
          'The expected email',
          email,
        ),
      )))
          .thenAnswer((_) => Future.value(loginResponse));
      final res = await client.mutate(
        MutationOptions(
          documentNode: document,
          variables: {'email': email, 'password': password},
        ),
      );
      expect(res, loginResponse);
    });
```

PR #827, opened 2021-03-04
Version update for downstream nullsafety usage

b-jan:
#### Fixes / Enhancements

- Updated dependencies to null safe versions
- Fixed some deprecated examples
- Updated CI so that `path: ../graphql` can be depended on in dev


b-jan:
Working locally on my side!

micimize (approved):

PR #881, opened 2021-04-27
asWatchQueryOptions now includes operationName

ol88:
### **Problem:**

Using the `Query` widget in graphql_flutter and passing QueryOptions with operationName specified and a document containing multiple queries result in error:
```
OperationException(linkException: null, graphqlErrors: [GraphQLError(message: Must provide a valid operation name if query contains multiple operations.
No operation name was given. , locations: null, path: null, extensions: null)])
```
### **Root cause:**
Currently both `QueryOptions` and `WatchQueryOptions` accept `operationName` in their constructors but `asWatchQueryOptions` method in `QueryOptions` currently does not passes this parameter to the `WatchQueryOptions` constructor resulting in `operationName == null` in the resulting `WatchQueryOptions` even if it was correctly specified in `QueryOptions`.

This does not happen when using `graphQLClient` directly since `WatchQueryOptions` is not used then. `graphQLClient`which works as expected and returns the correct query and results when operationName is specified with multiple query on the document.

### **Fix**
Add the missing paramater to `asWatchQueryOptions`:
```
  WatchQueryOptions asWatchQueryOptions({bool fetchResults = true}) =>
      WatchQueryOptions(
        document: document,
        operationName: operationName, //<- Add
        variables: variables,
```
The suggested fix works well for me and I don't believe it would cause any undesirable side-effects even if officially this package does not support multiple query per document, it works well with this.

### **Sample code:**
```

class MyClass extends StatefulWidget {
  @override
  _MyClassState createState() => _MyClassState();
}

class _MyClassState extends State<MyClass> {
  @override
    return Query(
      options: QueryOptions(
        document: gql(r'''
          query getData{
            data{
              __typename
              name
            }
          }

          query getOtherData{
            otherData{
              __typename
              name
            }
          }
        '''),
        operationName: 'getData',
        fetchPolicy: FetchPolicy.cacheAndNetwork,
        pollInterval: Duration(seconds: 20),
      ),
      builder: (QueryResult result,
          {VoidCallback? refetch, FetchMore? fetchMore}) {
        if (result.hasException) {
          print(result.exception.toString());
          return Text(result.exception.toString());
        }

        if (result.isLoading) {
          return Center(child: CircularProgressIndicator());
        }

        if (result.data != null) {
          print(result.data);
        }

        return Container();
      },
    );
  }
}
```


ol88:
The Mutation widget had the same issue and same fix but this time in the `MutationState` of the `Mutation` widget.

```
class MutationState extends State<Mutation> {

//...

  WatchQueryOptions get _providedOptions {
    final _options = WatchQueryOptions(
      document: widget.options.document,
      operationName: widget.options.operationName, //<- Add
      variables: widget.options.variables,
//...
```

micimize (approved):

kiloreux:
@HofmannZ Is this release on pub.dev yet? The latest release there doesn't contain this fix yet.

PR #896, opened 2021-05-18
Capture stack traces in UnknownExceptions

shyndman:
#### Fixes / Enhancements
- Added 
We've been running into exceptions that are pretty tricky to nail down without more context, and I was thinking that the inclusion of stack traces in UnknownExceptions would be helpful for us, and everybody else!

Related to https://github.com/nhost/nhost-dart/issues/9

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -180,7 +180,7 @@ workflows:
   nightly:
     triggers:
       - schedule:
-          cron: "23 * * * *"
+          cron: "0 12 * * *"
           filters:
             branches:
               only:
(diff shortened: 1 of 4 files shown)


micimize (approved):
god yes. This should really be done for all exceptions at the `gql` interface level imo

micimize (approved):

micimize (approved):

shyndman:
Thank you. :)

shyndman:
Version 5.0.0! YESSS!

(I'm actually pumped about this. No more warnings from pub!)

PR #952, opened 2021-08-10
docs(graphql_flutter): fix typo

jjangga0214:


mark-bromell:
> We would appreciate it if you could provide us with more info about this issue/pr!

The `GraphQLCleint` object in the readme is using `store` as a named parameter in its constructor, using `store` will cause an error because it does not exist. The named parameter that should be used is `cache`. This fix will clear any confusion (which I had) when reading the readme and following its code examples.

mark-bromell (approved):

vincenzopalazzo (approved):
LGTM

Good catch!

PR #1037, opened 2022-02-13
feat: Update options

budde377:
This PR improves equality checks on options by making options immutable and supports the `operator==`. This is used to decide if the queries should be re-fetched.

The PR contains:

* Mark all options as immutable with `@immutable`.
* New copy helpers used by #1034 and this PR to prevent updates on the options.
* Switches from `.equals` to `operator==` to compare equality


vincenzopalazzo (approved):
LGTM, also it is always better to introduce some refactoring to use some building blocks provided by the language.

Good Job!

PR #1038, opened 2022-02-14
fix: Read from cache first on cacheAndNetwork

budde377:
When fetch policy is `cacheAndNetwork`, we now return cache and then fetch from the network, if the cached result is available). If not we fetch from the network.

The current implementation always awaits the network response if available. But when the fetch policy is `cacheAndNetwork` we first want to return the cached result (if available) and then continue to fetch over the network in background. This PR fixes this.


vincenzopalazzo (approved):
Another good catch, this fixes a lot of problems inside the library, thanks to working on it.

vytautas-pranskunas-:
Thanks guys!

KubanychAsanov:
Good job!

PR #1044, opened 2022-02-17
fix: Fetch more is broken with `parserFn`

budde377:
It was raised in https://github.com/heftapp/graphql_codegen/issues/41 that the `fetchMore` throws an error when using `graphql_codegen` (i.e. `parserFn`). It is because we're creating the fetch-more `QueryOptions` without passing the `parserFn`. This PR fixes this.


vincenzopalazzo (approved):
I will make a new beta right now!

PR #1067, opened 2022-03-14
fix: Beta branch reliance

budde377:
This PR removes the checkout of `beta` branch, which we're not using anymore, and replaces it with a checkout of `main`.


vincenzopalazzo (approved):
good catch, my fault for this issue!

Thanks!

PR #1068, opened 2022-03-15
fix: Add .pubignore and dependency

budde377:
This PR adds tests to `.pubignore` to reduce package size and removes dependency override.


vincenzopalazzo (approved):
Good catch thanks!

PR #1074, opened 2022-03-19
fix: create results from options

budde377:
This PR changes how we create `QueryResults`. By always creating them from a given `BaseOption` we're effectively preserving type information. Losing type context has been the cause of two issues, latest #1072.

Since this PR changes the signature of the `QueryResult` which would warrant a major release if we had a better split of releases. This will cause issues for our beta users. The instruction should be to keep the packages in sync. Again, this is a major argument for persuing split releases.

### DIFF
--- a/.github/workflows/graphql_build.yml
+++ b/.github/workflows/graphql_build.yml
@@ -11,15 +11,6 @@ jobs:
         with:
           sdk: stable
       - name: Install dependencies
-        run: |
-          cd packages/graphql
-          dart pub get
+        run: make dep
       - name: Code formatting check
-        run: |
-          cd packages/graphql
-          dart format --set-exit-if-changed .
-      - name: Code analyze check
-        run: |
-          cd packages/graphql
-          dart analyze lib --fatal-infos
-          dart analyze test --fatal-infos
+        run: make ci_fmt_client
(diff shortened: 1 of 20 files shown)


budde377:
@vincenzopalazzo, with the breaking change mentioned above the current CI will always fail because some are using melos bootstrap and others aren't. What are your thoughts on this?

vincenzopalazzo:
>@vincenzopalazzo, with the breaking change mentioned above the current CI will always fail because some are using melos bootstrap and others aren't. What are your thoughts on this?

I will look into the error today, sorry for delay!

vincenzopalazzo (approved):
Good improvement, and sorry for the delay here but I have some trouble with Hive on flutter, basically the problem was that on clutter we use `Hive.openBox` without specifying the generics type and we specify in the graphql package the type.

This makes Hive a stupid software unable to recognize the status of the db.

But now it is fixed, and the CI should be up and running again!

apackin:
We're running into a type issue that I believe is from this change:

Screen Shot 2022-03-31 at 5 59 25 PM

vincenzopalazzo:
@apackin 

The version that you are using? In addition, a new issue with a good description of the problem can be very helpful, debugging in a screenshot is always frustrating for me, sorry!

PR #1147, opened 2022-05-24
fix: Export `useGraphQLClient` hook

### LINKED ISSUE #1146: graphql_flutter.dart not exporting useGraphQLClient hook?
According to [Other hooks](https://github.com/zino-hofmann/graphql-flutter/blob/main/packages/graphql_flutter/README.md#other-hooks) section of the README.md, it seems that a consumer should be able to use the `useGraphQLClient` hook.

However, after trying it myself, I could not make it work. 

You can reproduce it by 

1. Add graphql_flutter package import to your `HookWidget`

`import 'package:graphql_flutter/graphql_flutter.dart';`

2. In your widget's build method, try retrieving a reference to the GraphQL client by using the mentioned hook 
 
```dart
@override
   final graphQLClient = useGraphQLClient();
   ...
}
```

3. Expected behavior is for this to actually worked as documented. Actual behavior is this not compiling. 

Digging a bit on the source code, It looks like the only missing bit to export it as part of the [graphql_flutter.dart](https://github.com/zino-hofmann/graphql-flutter/blob/main/packages/graphql_flutter/lib/graphql_flutter.dart) barrel file.

Is this an intended behavior and the documentation went out-of-date? or should we consider this a bug?   

Cheers

budde377:
Fixes #1146

### DIFF
--- a/packages/graphql_flutter/lib/graphql_flutter.dart
+++ b/packages/graphql_flutter/lib/graphql_flutter.dart
@@ -14,5 +14,6 @@ export 'package:graphql_flutter/src/widgets/hooks/mutation.dart';
 export 'package:graphql_flutter/src/widgets/hooks/query.dart';
 export 'package:graphql_flutter/src/widgets/hooks/subscription.dart';
 export 'package:graphql_flutter/src/widgets/hooks/watch_query.dart';
+export 'package:graphql_flutter/src/widgets/hooks/graphql_client.dart';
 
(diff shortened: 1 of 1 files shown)


vincenzopalazzo (changes requested):
LGTM, just a small comment to the commit header message `fix(graphql_flutter)` instead of only `fix`

vincenzopalazzo (approved):

PR #1167, opened 2022-06-27
fix(graphql): Transform nested objects

budde377:
Recursively transforms nested objects #1091.


vincenzopalazzo (approved):
Great job! thanks!

I did not play with it, but I had some problems with the decoding in some of my other projects.

However, do we really need more than `Map.from()`? this usually makes the job, because it should work with iterator internally!

Maybe some weekend could be a good debug session on what happens, or fo you have already done this funny debug session?

PR #1238, opened 2022-09-29
Fix links in pubspec.yaml

ueman:
This fixes links in the pubspec.yaml files of this project.
As you can see in the diff, the current links still point to the master branch, which isn't used anymore. In addition, I've also added the issue tracker link.

#### Docs

- Updated links in pubspec.yaml files


vincenzopalazzo (approved):
thanks, just in time for the next beta

PR #1327, opened 2023-05-17
feat: bump sdk version upper bound to <4.0.0

ndelanou:
Make the package available for any Dart SDK versions before 4.0.0
Currently, the package doesn't work with the Dart beta channel (3.1.0)


vincenzopalazzo (approved):
Thanks!

PR #1394, opened 2023-11-25
fix(graphql): add missing response extensions

kamilkarp:
According to graphql specification: "The response map may also contain an entry with key `extensions`".

https://github.com/graphql/graphql-spec/blob/main/spec/Section%207%20--%20Response.md

Added `extensions` in the response.


vincenzopalazzo (approved):
LGMT

Thanks sorry for delay here

PR #1400, opened 2024-01-03
docs(graphql): replace `GraphQL` with `GraphQLClient`

mahmoud-eslami:
### Description
There is a typo in [GraphQL documentation](https://github.com/zino-hofmann/graphql-flutter/blob/main/packages/graphql/README.md), it seems the return type should be `GraphQLClient` instead of `GraphQL`.


vincenzopalazzo (approved):
LGTM, good catch thanks!

PR #1404, opened 2024-02-03
Main

szantogab:
There was an issue with the `WebsocketLink`, because it parses every websocket message multiple times, unnecessarily slowing down the app. In fact it was so slow that my app was unusable.

So I made the `messages` stream a `broadcast stream`, this fixes the issue.

[1](https://github.com/zino-hofmann/graphql-flutter/blob/main/docs/dev/MAINTAINERS.md#commit-style)

<!--
### Breaking changes

No breaking changes.


vincenzopalazzo (approved):
LGTM

Sorry for delay here I miss this PR

PR #1451, opened 2024-08-04
feat(graphql): Configurable option to deduplicate pollers

kvenn:
### Problem
If you call startPolling 10 times for the same request, it'll start 10 pollers that are all doing the same thing.

### Solution
* Allow consumers to specify if they'd like deduplication (defaults to false for backwards compatibility)
* When a new poller is added, if it's a shorter duration than the previous one for that identical request, it'll update to use the new time
	* If that faster poller is removed, it'll fallback to the next fastest time
	* If all pollers for a request are removed, it doesn't poll

### Testing
```dart
    final obs5 = _client.friendRequests(_onMyFriendRequests);
    final obs10 = _client.friendRequests(_onMyFriendRequests);
    final obs3 = _client.friendRequests(_onMyFriendRequests);
    final obs15 = _client.friendRequests(_onMyFriendRequests);

    obs5.startPolling(const Duration(seconds: 5));
    obs10.startPolling(const Duration(seconds: 10));
    obs3.startPolling(const Duration(seconds: 3));
    obs15.startPolling(const Duration(seconds: 15));
    // Should settle at 3 seconds (fastest)

    // Wait 10 seconds to confirm pollers only every 3 seconds
    Future.delayed(const Duration(seconds: 7)).then((_) {
      print('------------------TEST------------------');
      obs3.stopPolling();
      // Duration should now be 5 seconds
    });
    // Confirm it's 5 (should see 3 or 4 of them)
    Future.delayed(const Duration(seconds: 20)).then((_) {
      print('------------------TEST------------------');
      obs5.close();
      // Duration should now be 10 seconds
    });
```

### DIFF
--- a/packages/graphql/lib/src/core/query_manager.dart
+++ b/packages/graphql/lib/src/core/query_manager.dart
@@ -27,6 +27,7 @@ class QueryManager {
     required this.link,
     required this.cache,
     this.alwaysRebroadcast = false,
+    bool deduplicatePollers = false,
   }) {
     scheduler = QueryScheduler(
       queryManager: this,
--- a/packages/graphql/lib/src/core/query_manager.dart
+++ b/packages/graphql/lib/src/core/query_manager.dart
@@ -31,6 +31,7 @@ class QueryManager {
   }) {
     scheduler = QueryScheduler(
       queryManager: this,
+      deduplicatePollers: deduplicatePollers,
     );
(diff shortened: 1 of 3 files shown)


vincenzopalazzo (approved):
LGTM, would be good to add some kind of docs but this can be a followup PR

PR #1452, opened 2024-08-04
feat(graphql): Reuse request results

kvenn:
### Problem
If you have two observables for the same request, it will unnecessarily do extra (expensive) work
* `maybeRebroadcast` will diff, denormalize, and then require parsing again
	* This is more substantial since this gets fired on every network response and cache write
* `fetch(policy.cache)` will need to denormalize and require parsing again
	* This will cause a flash of delay when attaching to an observable that already has data for it
* ex: two observables of the GetUser query for the same resource (id: `xyz`)

### Solution
I went with a solution that doesn't require changing any core flows or data structures. Slightly based on how Ferry keeps a [map of requests](https://github.com/gql-dart/ferry/blob/master/packages/ferry/lib/src/request_controller_typed_link.dart#L34).

The more observables you have for the same `Request`s, the bigger improvements you will see

* `_getQueryResultByRequest` - Instead of introducing a new cache, we iterate through the `queries` array to find if any queries match the `Request` 
* When we `maybeRebroadcast` of `fetch` from cache, we check this first
* The new `copyWith` on `QueryResult` lets us re-use the bulk of the existing query result (including the parsed data)

Other improvement added - call maybeRebroadcast less
* Don't call it on failures (since the cached data didn't change)

### DIFF
--- a/packages/graphql/lib/src/cache/cache.dart
+++ b/packages/graphql/lib/src/cache/cache.dart
@@ -204,9 +204,23 @@ class GraphQLCache extends NormalizingDataProxy {
   ///
   /// This allows for hierarchical optimism that is automatically cleaned up
   /// without having to tightly couple optimistic changes
+  ///
+  /// This is called on every network result as cleanup
   void removeOptimisticPatch(String removeId) {
+    final patchesToRemove = optimisticPatches
+        .where(
+          (patch) =>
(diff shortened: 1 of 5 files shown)


vincenzopalazzo (approved):
LGTM, thanks

PR #1462, opened 2024-10-09
Update dependencies including Web socket channel 3.0.1

redDwarf03:
Update dependencies including Web socket channel 3.0.1 from graphql_flutter 5.1.2

Note that I've removed the examples for null safety reasons. I didn't see the point of updating in view of the new version that will be released.


Chralu:
Careful, automated tests are broken here.

It seems that we do not wait for connection to be not well established before moving on.

```log
Building package executable... 
Built test:test.
00:01 +65 -1: test/websocket_test.dart: SocketClient without payload disconnect via dispose [E]                                                       
  Expected: not null
    Actual: <null>
  
  package:matcher                 expect
  test/websocket_test.dart 178:7  main.<fn>.<fn>
  

To run this test again: /Users/charly/.asdf/installs/flutter/3.24.3-stable/bin/cache/dart-sdk/bin/dart test test/websocket_test.dart -p vm --plain-name 'SocketClient without payload disconnect via dispose'
00:01 +70 -2: test/websocket_test.dart: SocketClient without payload graphql-transport-ws disconnect via dispose graphql-transport-ws [E]             
  Expected: not null
    Actual: <null>
  
  package:matcher                 expect
  test/websocket_test.dart 442:7  main.<fn>.<fn>
  

To run this test again: /Users/charly/.asdf/installs/flutter/3.24.3-stable/bin/cache/dart-sdk/bin/dart test test/websocket_test.dart -p vm --plain-name 'SocketClient without payload graphql-transport-ws disconnect via dispose graphql-transport-ws'
00:04 +76 -2: Some tests failed.                                                                                                                      

Consider enabling the flag chain-stack-traces to receive more detailed exceptions.
For example, 'dart test --chain-stack-traces'.
```

PR #1475, opened 2024-12-19
feat(graphql) make timeout configurable in query options

ndelanou:
### Breaking changes
none

#### Fixes / Enhancements

- Allow to specify a `queryRequestTimeout` in every option classes that extends `BaseOptions`. This new parameter overrides the `queryRequestTimeout` set in the `GraphQLClient` constructor.
- The `queryRequestTimeout` parameter from `GraphQLClient` constructor is now nullable. This will allow having no timeout on queries (#1473)

### DIFF
--- a/packages/graphql/lib/src/core/_base_options.dart
+++ b/packages/graphql/lib/src/core/_base_options.dart
@@ -23,6 +23,7 @@ abstract class BaseOptions<TParsed extends Object?> {
     ErrorPolicy? errorPolicy,
     CacheRereadPolicy? cacheRereadPolicy,
     this.optimisticResult,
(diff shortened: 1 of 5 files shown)


vincenzopalazzo (commented):
Thanks, can you squash the commit https://github.com/zino-hofmann/graphql-flutter/pull/1475/commits/141350ad5466d8f3d908cff16df440a9d415a277 inside the previous one?

In additioncan you follow the commit guideline that we are inside the contributor guide?

Otherwise LGMT

ndelanou:
@vincenzopalazzo should be better now 👍

PR #1510, opened 2025-12-03
Feature/add cancellation support

lewinpauli:
As needed since 6 years here is cancellation support [https://github.com/zino-hofmann/graphql-flutter/issues/181](https://github.com/zino-hofmann/graphql-flutter/issues/181) 

- Add CancellationToken class for operation cancellation
- Add CancelledException for cancelled operations
- Integrate cancellation into BaseOptions, QueryOptions, MutationOptions
- Implement cancellation handling in QueryManager
- Add queryCancellable() and mutateCancellable() convenience methods
- Add comprehensive test suite (7 tests)
- Add usage documentation

-> no breaking changes

-> all tests pass


vincenzopalazzo (changes requested):
Thanks for the contribution and for tackling #181, this is a feature that has been requested for a long time!

However, there are some critical issues that need to be addressed before this can be merged:

**Critical**

1. The custom `HttpLink` shadows the upstream one from `gql_http_link` and **breaks file upload support** (multipart requests). The current implementation only does `json.encode(body)`, so `GraphQLUpload` won't work anymore. This is a breaking change despite the PR description saying otherwise.

2. `io_client.dart` creates a new `dart:io` `HttpClient` for every request. This destroys connection pooling, DNS caching, and TLS session reuse — it will cause noticeable performance regression and socket exhaustion under load.

3. `io_client.dart` has a completion bug: the code uses both a `Completer` and direct `throw` statements. When exceptions are thrown (e.g. lines checking `isCancelled`), they bypass the `completer` but the method returns `completer.future` at the end, so those throws won't reach the caller.

4. The conditional import uses `dart.library.html` which is deprecated — this will break on WASM targets. Use `dart.library.js_interop` instead.

**Cleanup needed**

5. Remove `examples/starwars/ios/Flutter/ephemeral/flutter_lldb_helper.py` and `flutter_lldbinit` — these are generated build artifacts.

6. Remove the `CLAUDE.md` from this PR, it's unrelated to cancellation support.

7. `mutateCancellable` manually reconstructs every `MutationOptions` field instead of using a `copyWith` — this is fragile and will silently drop new fields. Add a proper `copyWith` to `MutationOptions` like `QueryOptions` has.

8. Error detection by string matching (`e.message.contains('abort')`) in `http_link.dart` is fragile and will silently swallow legitimate errors.

**Suggestion**

Consider adding cancellation at the `QueryManager` level only (which the PR partially does already) without replacing `HttpLink`. The HTTP-level abort could live in a separate `CancellableHttpLink` that extends the original rather than shadowing it. This keeps file uploads working and avoids the `gql_http_link` compatibility issues.

for commit style see https://github.com/zino-hofmann/graphql-flutter/blob/main/docs/dev/MAINTAINERS.md#commit-style

lewinpauli:
Thank you for the detailed review and the clear todos @vincenzopalazzo 

I will have time for that in april

vincenzopalazzo:
Thank @lewinpauli do you have also time to look if this fix your problem? https://github.com/zino-hofmann/graphql-flutter/pull/1518 but anyway wait till you are ready it is fine

vincenzopalazzo (changes requested):
## Code Review

### Positive Aspects
- Good work addressing the previous review feedback: custom HttpLink shadow removed, `CancellableHttpLink` introduced as a separate class, `copyWith` added to `MutationOptions`, build artifacts cleaned up.
- The `CancellationToken` / `CancellableOperation` API is clean and idiomatic for Dart.
- Test coverage is solid — both unit tests with mock links and integration tests with real HTTP servers.

---

### Issues & Suggestions

#### 1. `QueryOptions.copyWithOptions` silently drops `cancellationToken` (Critical)

`queryCancellable` in `graphql_client.dart` calls:
```dart
final modifiedOptions = options.copyWithOptions(
  cancellationToken: cancellationToken,
);
```

But looking at the diff, `cancellationToken` is only being added to the `copyWith` method, **not** to `copyWithOptions`. If this actually compiles it means the named param was added but not plumbed through — every `copyWithOptions` / `copyWithPolicies` / `withFetchMoreOptions` / `asWatchQueryOptions` / `copyWithFetchPolicy` / `copyWithPollInterval` / `copyWithVariables` / `copyWithOptimisticResult` call that doesn't forward `cancellationToken` will silently **drop** the token.

The diff shows `cancellationToken` being threaded through `copyWithPolicies` and `asWatchQueryOptions`, which is good. But please verify that ALL copy/clone paths on `QueryOptions` and `WatchQueryOptions` forward `cancellationToken`. Missing it on any one path means the token silently disappears mid-pipeline.

#### 2. `CancellableOperation` never disposes its token (Medium)

`queryCancellable` and `mutateCancellable` both create a `CancellationToken` internally but **never dispose it**. The token holds a `StreamController.broadcast()` — if the caller only calls `cancel()` and never `dispose()`, that controller leaks.

The docs tell users "Don't forget to dispose the token when done" for the manual path, but the convenience methods hide the token. Consider either:
- Auto-disposing the token when the result future completes (via `.whenComplete`)
- Or documenting that `CancellableOperation.cancel()` also disposes

#### 3. Cancellation detection by string matching (Medium)

`CancellableHttpLink.request()` does:
```dart
if (error is http.ClientException &&
    (error.message.contains('cancelled') ||
        error.message.contains('abort')))
```

This is fragile. A legitimate `ClientException` with "Connection aborted by server" would be incorrectly swallowed as a `CancelledException`. The PR docs acknowledge this limitation, but a better approach would be to check `cancellationToken.isCancelled` before converting:
```dart
if (cancellationToken?.isCancelled == true &&
    error is http.ClientException) {
  // Only convert to CancelledException if WE actually cancelled
}
```

This eliminates false positives entirely.

#### 4. `_cancellable_io_client.dart` — Completer + direct return mix (Medium)

The IO client uses a `Completer<http.Response>` but also has multiple `return completer.future` exit paths scattered through cancellation checks (lines ~1155, ~1200, ~1225). The method signature returns `Future<http.Response>` and the final `return completer.future` at the bottom catches the normal path, but the intermediate returns create a confusing flow.

More importantly: when `isCancelled` is detected in the `await for (final chunk in httpClientResponse)` loop (line ~1222), it returns `completer.future` — but at that point the `cancellationSubscription` listener may have already completed the completer with an error, so this return races with the listener. The `finally` block then tries to cancel the subscription regardless.

Consider simplifying: just throw on cancellation checks instead of completing a completer, since the outer try/catch already handles it.

#### 5. Multipart request `finalize()` called twice (Medium)

In `_cancellable_io_client.dart` for `http.MultipartRequest` handling (line ~1175):
```dart
final stream = request.finalize();
```

`MultipartRequest.finalize()` can only be called once — it marks the request as finalized. If any retry logic or error recovery re-enters this code path with the same request object, it will throw `StateError("Can't finalize a finalized MultipartRequest")`. This is a concern because the `CancellableHttpLink._prepareRequest` creates the `http.MultipartRequest` once, and if the stream is replayed (Link retry, etc.), the same request object would hit `finalize()` again.

Is there any path where a link chain could retry the same request? If so, this needs a factory pattern.

#### 6. `web` dependency added but only used by `CancellableHttpLink` (Cleanup)

The `web: ^1.1.1` dependency is added to the core `graphql` package `pubspec.yaml`. This pulls `package:web` (and transitively `dart:js_interop`) into every consumer. Since `CancellableHttpLink` is opt-in and only needed when users want HTTP-level cancellation, consider whether this should live in a separate package (e.g., `graphql_cancellable_http_link`) to avoid bloating the dependency tree for everyone.

At minimum, the conditional import `if (dart.library.js_interop)` should ensure the web code is tree-shaken on non-web targets, but the dependency still shows up in `pubspec.yaml` resolution for all platforms.

#### 7. `CancelledException` constructor is not `const` (Cleanup)

```dart
class CancelledException extends LinkException {
  CancelledException(this.message) : super(null, null);
```

The class is `@immutable` but the constructor isn't `const`. This is fine functionally but inconsistent with the annotation. Either make it `const` (if `LinkException` supports it) or drop `@immutable`.

#### 8. `onDone` only cancels `cancellationSubscription`, not `subscription` (Cleanup)

In `query_manager.dart` (line ~774):
```dart
onDone: () {
  cancellationSubscription?.cancel();
},
```

When the stream completes normally (`onDone`), only the cancellation subscription is cancelled. The response stream subscription itself should also be cleaned up here, though the `finally` block handles it. This is just redundant / asymmetric — consider removing the `onDone` callback entirely since `finally` does the cleanup.

#### 9. Excessive documentation duplication (Cleanup)

`CANCELLATION_USAGE.md` (312 lines) and the additions to `packages/graphql/README.md` (132 lines) contain heavily overlapping content. The README already covers the full API — `CANCELLATION_USAGE.md` adds lengthy examples that duplicate what's in the README.

Consider keeping only the README section (which is where users look) and removing `CANCELLATION_USAGE.md`, or trimming `CANCELLATION_USAGE.md` down to a brief "advanced usage" addendum.

---

### Summary

Concept ACK — the cancellation API design is sound and the previous review feedback was well addressed. The main concerns are: token disposal leak in convenience methods, string-matching cancellation detection (should check `isCancelled` instead), and the completer/return flow in the IO client. Fix those and this is merge-ready.

lewinpauli:
Update: the link now delegates cancellation to the standard http package (v1.6.0+) instead of custom clients.

- CancellableHttpLink builds an abortTrigger future from the CancellationToken and passes it to http.AbortableRequest / http.AbortableMultipartRequest

- http.Client handles the network-level abort directly, so we get true HTTP cancellation on all platforms (including web) from a single code path

- Removed the custom _cancellable_client_interface.dart, _cancellable_io_client.dart, and _cancellable_web_client.dart — no platform shims anymore

- Dropped the web dependency; bumped http to ^1.6.0

- Added a test that verifies the underlying connection is actually aborted (not just the Dart future short-circuiting)

Net result: cancelled requests now appear as (canceled) in DevTools/native network traces, and file uploads stop streaming immediately on cancel.

lewinpauli:
Hi,

can someone review this pls?

@jorg1piano @vincenzopalazzo

examples/starwars/ios/Flutter/ephemeral/flutter_lldb_helper.py
```diff
@@ -0,0 +1,32 @@
+#
```
vincenzopalazzo: **Medium:** Generated build artifact — should not be committed.

This was flagged in the first review and commit `3186b30` says "Remove generated build artifacts" but they reappeared in later commits. Can we delete these from the PR?

packages/graphql/lib/src/links/links.dart
```diff
@@ -2,3 +2,5 @@
 export 'package:graphql/src/links/gql_links.dart';
 export 'package:graphql/src/links/auth_link.dart';
 export 'package:graphql/src/links/websocket_link/websocket_link.dart';
+export 'package:graphql/src/links/cancellable_http_link/cancellable_http_link.dart'
```
vincenzopalazzo: **Cleanup:** Duplicate export — `gql_links.dart` already exports `cancellable_http_link.dart` (lines 5-6), and this file re-exports `gql_links.dart` on line 2. So this direct export is redundant. Non-blocking nit.

vincenzopalazzo (commented):
One issue to fix before merge: generated build artifacts are still in the diff.

## Code Review

### Positive Aspects
- Good work addressing all the feedback from the previous two rounds: custom IO/web clients removed in favor of `http` package's `AbortableRequest`, token auto-disposal via `whenComplete`, `isCancelled` check instead of string matching, `CancelledException` made `const`, duplicate docs removed.
- The refactor to `http.AbortableRequest` (http >=1.6.0) is the right call — single code path, no conditional imports, proper connection pooling. The `http` package handles platform differences internally.
- `cancellationToken` correctly forwarded through ALL copy/clone paths: `copyWithOptions`, `withFetchMoreOptions`, `asWatchQueryOptions`, `copyWithPolicies`, and all `WatchQueryOptions`/`MutationOptions` variants. Verified each one.
- Solid test coverage: 7 unit tests with mock links + 13 integration tests with real HTTP servers covering cancellation, file uploads, pre-cancelled tokens, server errors, concurrent requests, and connection reuse.

---

### Issues & Suggestions

#### 1. Generated build artifacts still in the PR (Medium)

`examples/starwars/ios/Flutter/ephemeral/flutter_lldb_helper.py` and `flutter_lldbinit` are generated Xcode/Flutter build artifacts that should not be committed. This was flagged in the first review and commit `3186b30` says "Remove generated build artifacts" — but they're still showing as additions in the final diff. Can we remove them?

#### 2. Duplicate export of `CancellableHttpLink` (Cleanup)

`gql_links.dart:5-6` exports `cancellable_http_link.dart`, and `links.dart:5-6` also exports it directly. Since `links.dart` already re-exports `gql_links.dart` (line 2), `CancellableHttpLink` is effectively exported twice through `links.dart`. Non-blocking, but can we remove the redundant one to keep the export graph clean?

---

### Summary

Concept ACK on the architecture and implementation. The `AbortableRequest` approach is the right call. The cancellation API is well-designed and idiomatic for Dart.

Remove `flutter_lldb_helper.py` and `flutter_lldbinit`, and this is ready to merge. The duplicate export is non-blocking cleanup.

Targeted checks on the PR head passed locally: `dart analyze packages/graphql/` (no issues), `dart test` (113/113 pass).

examples/starwars/ios/Flutter/ephemeral/flutter_lldb_helper.py
lewinpauli: fixed in 6cb60136f89884ca5c7f1c024c125be5e00ccf5a

packages/graphql/lib/src/links/links.dart
lewinpauli: fixed in 6cb60136f89884ca5c7f1c024c125be5e00ccf5a

vincenzopalazzo (commented):
Concept ACK — the code is safe to merge. The two previous review rounds were well addressed (build artifacts removed + gitignored, duplicate export dropped, `web` dep replaced with `http` `AbortableRequest`, `CancelledException` made `const`, token forwarded through all copy paths). No blocking correctness issues remain. Two preconditions before merge: **(1) `/git-cleanup` the commits to follow the repo commit style, and (2) a small docstring fix.** Details below.

## Code Review

### Positive Aspects
- The refactor to `http.AbortableRequest`/`AbortableMultipartRequest` (http >=1.6.0) is the right call — single code path, no conditional imports, real network-level abort on all platforms, and `_client` is a shared `http.Client` so connection pooling/DNS/TLS reuse is preserved (verified by the "multiple requests reuse connection" test).
- Cancellation detection no longer relies on string matching — `error is http.RequestAbortedException || (cancellationToken?.isCancelled == true && error is http.ClientException)` in `cancellable_http_link.dart:144-146` eliminates the false-positive swallowing I flagged earlier.
- `cancellationToken` is correctly threaded through **every** copy/clone path on `QueryOptions`, `WatchQueryOptions`, `SubscriptionOptions`, and `MutationOptions` (`copyWithOptions`, `withFetchMoreOptions`, `asWatchQueryOptions`, `copyWithPolicies`, `copyWithFetchPolicy`, `copyWithPollInterval`, `copyWithVariables`, `copyWithOptimisticResult`). Re-verified each one — no silent drop.
- Solid coverage: 20 cancellation tests (mock-link unit tests + real-HTTP-server integration tests covering abort, file upload, pre-cancelled token, server-error passthrough, concurrent cancel, connection reuse).

Targeted checks I ran on the PR head (`1e7a060`): `dart analyze lib test` → no issues; `dart test` → 113/113 pass; `dart test test/cancellation_test.dart test/cancellable_http_link_test.dart` → 20/20 pass; `dart format --line-length=80` → clean.

---

### Issues & Suggestions

#### 1. `queryCancellable`/`mutateCancellable` docstrings show a `try/catch` that will never fire (Medium)

`graphql_client.dart:213-219` and `253-260` document the result like this:
```dart
/// try {
///   final result = await operation.result;
/// } catch (e) {
///   if (e is CancelledException) { ... }
/// }
```
But the actual (and tested) behavior is that cancellation does **not** throw — `QueryManager._resolveQueryOnNetwork` catches the `CancelledException` and delivers it via `result.exception.linkException` (see `cancellation_test.dart:50-52`: `expect(result.exception!.linkException, isA<CancelledException>())`). A user following the docstring's `try/catch` would never catch anything. The `CancellationToken` docstring (`cancellation_token.dart:26-28`) gets this right — "will complete with a QueryResult containing a CancelledException". Can we align the two `graphql_client.dart` docstrings with the real behavior? This is the one thing I'd want fixed before merge since it's user-facing API docs.

#### 2. `MultipartFile.finalize()` is single-use — confirm no retry link in the chain (Medium, documented limitation)

`cancellable_http_link.dart:281` calls `f.finalize()` while building the multipart body, and the method doc (lines 211-215) correctly warns that a finalized file can't be reused. Just want to confirm: is there any path where a retry/error link sitting in front of `CancellableHttpLink` could replay the **same** `Request` and re-enter `_prepareRequest` on already-finalized files? If `RetryLink`-style usage is supported, this throws `StateError`. The doc note is good; non-blocking as long as we're agreed the contract is "retry must re-run the full link chain, not replay the BaseRequest".

#### 3. `_createAbortTrigger` can leave a dangling `StateError` when the token is disposed (Cleanup)

`cancellable_http_link.dart:299-303` uses `token.onCancel.first` as the abort trigger. On a normal (non-cancelled) completion, `CancellableOperation` disposes the token via `whenComplete`, which closes the broadcast controller **without** emitting — so the `.first` future completes with `StateError("No element")`. It's currently absorbed by `runZonedGuarded` (the controller is already closed, so line 172 silently ignores it), so there's no crash. But relying on the zone guard to swallow a guaranteed error on every successful request is subtle. Consider driving the trigger from a dedicated `Completer` that you only complete on cancel, so the happy path produces no error at all. Non-blocking.

#### 4. `CancellableHttpLink` is a hand-rolled fork of `gql_http_link`'s `HttpLink` (~430 lines) (Suggestion)

This duplicates header handling, GET-for-queries encoding, multipart mapping, and response parsing from upstream `HttpLink`. It works, but it will drift — bug fixes/behavior changes in `gql_http_link.HttpLink` won't propagate here. Worth either a `// keep in sync with gql_http_link HttpLink` marker, or a follow-up to upstream `abortTrigger` support into `gql_http_link` so we can delete this fork. Not for this PR.

---

### Commit history — required before merge (your `/git-cleanup` question)

Good news: `origin/main` is already an ancestor of the branch, so `git rebase -i origin/main` is **conflict-free** — `/git-cleanup` will be clean. But the current 16 commits + 1 merge commit do **not** follow `docs/dev/MAINTAINERS.md#commit-style` and must be cleaned up first. Concretely, the cleanup needs to:

- **Squash the fixups**: 6 pure `style(graphql): apply dart format` / `dart format --line-length=80 ./` commits, plus the `fix(graphql): address … review feedback` commits, are all fixups of the feature and should fold into it.
- **Drop the merge commit**: `chore(graphql): merge origin/main into feature branch` — `chore` isn't an allowed type and the rebase flattens it anyway.
- **Fix the types/scopes** — the repo only allows types `feat`/`fix`/`deprecate`/`remove` and scopes `graphql`/`graphql_flutter`. Current violations: `feature:` (b86110e), `refactor:` (7335aac), `style:` (×6), `chore:` (merge), `feat(demo):` (42d3001 — `demo` isn't a valid scope, use `graphql_flutter`), and several untyped commits (`Added true HTTP request cancellation…` is also past-tense, capitalized, with a trailing `.`).
- **Add a closing footer** — the PR body links #181 but without a closing keyword, so the changelog won't auto-link. Add `Closes #181`.

Recommended end state (1–2 commits):
```
feat(graphql): add cancellation support for queries and mutations

Add CancellationToken/CancellableOperation, CancelledException, and the
opt-in CancellableHttpLink (http AbortableRequest) for true network-level
abort. Wire cancellationToken through QueryManager and all options copy
paths; add queryCancellable()/mutateCancellable() convenience methods.

Closes #181
```
and, if you want the example separate:
```
feat(graphql_flutter): add cancellation example with local test server
```

---

### Summary

Concept ACK — implementation is merge-safe (analyze clean, 113+20 tests green, no rebase conflicts against current main). **Two gates before merge:** run `/git-cleanup` to produce conventional commits per `MAINTAINERS.md` (squash the format/fixup commits, drop the merge commit, fix types/scopes, add `Closes #181`), and fix the misleading `queryCancellable`/`mutateCancellable` docstrings (#1). Items #2–#4 are non-blocking. Nice work pushing this 6-year-old request across the line!

packages/graphql/lib/src/graphql_client.dart
```diff
@@ -191,6 +191,89 @@ class GraphQLClient implements GraphQLDataProxy {
...
+  /// operation.cancel();
+  ///
+  /// // The result future will complete with a CancelledException
+  /// try {
+  ///   final result = await operation.result;
+  /// } catch (e) {
+  ///   if (e is CancelledException) {
+  ///     // Handle cancellation
+  ///   }
+  /// }
```
vincenzopalazzo, line 216: **Medium:** This `try/catch` will never fire. Cancellation does not throw — `QueryManager._resolveQueryOnNetwork` catches the `CancelledException` and delivers it via `result.exception.linkException` (asserted in `cancellation_test.dart:50-52`). A user copying this snippet would never catch anything. Please align with the actual behavior, e.g.:

```suggestion
  /// // The result future completes with a QueryResult whose exception is set
  /// final result = await operation.result;
  /// if (result.exception?.linkException is CancelledException) {
  ///   // Handle cancellation
  /// }
```

packages/graphql/lib/src/graphql_client.dart
```diff
@@ -191,6 +191,89 @@ class GraphQLClient implements GraphQLDataProxy {
...
```
vincenzopalazzo, line 254: **Medium:** Same as the `queryCancellable` docstring — cancellation is delivered in `result.exception.linkException`, not thrown, so this `try/catch` is misleading. Please mirror the corrected snippet here too.

packages/graphql/lib/src/links/cancellable_http_link/cancellable_http_link.dart
```diff
@@ -0,0 +1,427 @@
...
+      final fileMapping = <String, List<String>>{};
+      final fileEntries = fileMap.entries.toList();
+      for (var i = 0; i < fileEntries.length; i++) {
+        final entry = fileEntries[i];
+        final indexString = i.toString();
+        fileMapping[indexString] = [entry.key];
+        final f = entry.value;
+        multipartRequest.files.add(http.MultipartFile(
+          indexString,
+          f.finalize(),
```
vincenzopalazzo, line 289: **Medium (documented limitation):** `finalize()` is single-use — if a retry/error link in front of `CancellableHttpLink` ever replays the same `Request` and re-enters `_prepareRequest` on already-finalized files, this throws `StateError`. The method doc (lines 211-215) already calls this out; just confirming the agreed contract is "a retry re-runs the full link chain, not the same `BaseRequest`." Non-blocking.

packages/graphql/lib/src/links/cancellable_http_link/cancellable_http_link.dart
```diff
@@ -0,0 +1,427 @@
...
+      ..body = httpBody
+      ..headers.addAll(headers);
+  }
+
+  /// Creates a [Future] that completes when the [cancellationToken] is
+  /// cancelled, suitable for passing as [http.AbortableRequest.abortTrigger].
+  static Future<void>? _createAbortTrigger(CancellationToken? token) {
+    if (token == null) return null;
+    if (token.isCancelled) return Future<void>.value();
+    return token.onCancel.first;
```
vincenzopalazzo: **Cleanup:** `token.onCancel.first` completes with `StateError("No element")` on every *non-cancelled* request, because `CancellableOperation` disposes the token (closing the broadcast controller without emitting) once the result completes. It's currently swallowed by `runZonedGuarded` since the controller is already closed, so no crash — but relying on the zone guard to absorb a guaranteed error on the happy path is subtle. Consider a dedicated `Completer<void>` completed only on cancel, so success produces no dangling error. Non-blocking.

packages/graphql/lib/src/graphql_client.dart
lewinpauli: Good catch fixed it. Updated the docstring to match the actual (tested) behavior: cancellation isn't thrown, it's delivered on the resolved QueryResult via result.exception?.linkException. Applied your suggested snippet verbatim:

Resolved in the latest push.

lewinpauli: Good catch — fixed. Updated the docstring to match the actual (tested) behavior: cancellation isn't thrown, it's delivered on the resolved QueryResult via result.exception?.linkException. Applied your suggested snippet verbatim:

```
// The result future completes with a QueryResult whose exception is set
final result = await operation.result;
if (result.exception?.linkException is CancelledException) {
  // Handle cancellation
}
```

Resolved in the latest push.

packages/graphql/lib/src/graphql_client.dart
lewinpauli: Fixed here too, mirrored the same corrected result.exception?.linkException is CancelledException snippet so both convenience methods document the real delivery mechanism. Resolved in the latest push.

packages/graphql/lib/src/links/cancellable_http_link/cancellable_http_link.dart
lewinpauli: Confirmed, that's exactly the agreed contract. A retry must re-run the full Link chain (which re-enters request() → _prepareRequest and builds fresh MultipartFile objects), not replay the same http.BaseRequest. This matches how RetryLink/error links operate in gql_link, and it's the same finalize-on-send model the upstream HttpLink uses. I've left the method doc note in place so the limitation stays documented. No code change needed.

packages/graphql/lib/src/links/cancellable_http_link/cancellable_http_link.dart
lewinpauli: Fixed as suggested. Replaced token.onCancel.first with a dedicated Completer<void> plus a stream subscription, the completer only completes when the token actually fires; on normal completion (token disposed → controller closes without emitting), the onDone handler just cancels the subscription and leaves the future uncompleted. So the happy path no longer produces a StateError for runZonedGuarded to swallow, and a never-completing trigger simply means "never aborted," which is the desired behavior. Resolved in the latest push.

lewinpauli:
Thanks for the thorough review @vincenzopalazzo ! Pushed a follow-up addressing all four items:

Docstrings (#1) queryCancellable/mutateCancellable now document cancellation as delivered via result.exception?.linkException instead of a try/catch that never fires.
_createAbortTrigger (#3) now uses a dedicated Completer<void> completed only on cancel; no more guaranteed StateError on the happy path.
Fork drift (#4) added a "keep in sync with gql_http_link HttpLink" maintenance note to the CancellableHttpLink doc, flagging it as a deliberate mirror until abortTrigger can be upstreamed.
finalize() (#2) confirmed the retry contract (re-run the chain, don't replay the BaseRequest); no change needed.
Also ran /git-cleanup: the branch is now two conventional commits per MAINTAINERS.md (feat(graphql): … with Closes #181, and feat(graphql_flutter): add cancellation example with local test server), with the format/fixup commits squashed and the merge commit dropped.

Validation on the graphql package: dart analyze → no issues; dart test → 113/113 pass (incl. 20 cancellation tests). (The examples/starwars analyzer errors are pre-existing on main, that demo is pinned to the old published graphql-5.2.4.)

vincenzopalazzo:
This need a rebase before another review! thanks

lewinpauli:
Rebased onto latest main. The conflict was in _resolveQueryOnNetwork, where #your cancellation work and the new timeout guard from 91e50cf touched the same block. Merged both: kept the guarded single-completer/subscription with the onDone StateError handling, and layered cancellation on top (pre-subscribe isCancelled check + an onCancel listener that settles the completer with CancelledException and cancels the response subscription). The no-token path is now identical to upstream, so the timeout behavior is unchanged. dart analyze clean, dart test green at 114 (94 upstream + 20 cancellation).

lewinpauli:
@vincenzopalazzo can you review

packages/graphql/test/cancellable_http_link_test.dart
```diff
@@ -0,0 +1,761 @@
...
+        // This test mainly ensures the flow works without hanging
+
+        cancellationToken.dispose();
+        await link.dispose();
+      } finally {
+        await slowServer.close(force: true);
+      }
+    });
+
+    test('multiple concurrent requests - cancel one, others complete',
```
vincenzopalazzo, line 368: **Medium (non-blocking):** Flaky under parallel load — this test fails intermittently in the full suite (the non-cancelled request assertion at `:419` gets `hasException == true`), but passes 3/3 in isolation.

The real `HttpServer` + hardcoded 2s delay races with the other tests binding servers concurrently. Suggest either tagging this group `@Tags(['integration'])` to run serially, or making it deterministic with a mock link.

vincenzopalazzo (commented):
Concept ACK — both pre-merge gates from the last round are satisfied. The squashed history is clean (2 commits, conventional types/scopes, `Closes #181`), and the `queryCancellable`/`mutateCancellable` docstrings now correctly show `result.exception?.linkException is CancelledException` instead of the misleading `try/catch`. This is merge-ready.

Production Readiness: GO

## Code Review

### Positive Aspects
- Commit history is now clean — exactly the 2-commit end state I recommended (`feat(graphql)` + `feat(graphql_flutter)`) with the `Closes #181` footer so #181 auto-links. The 16 commits + merge commit are gone.
- Docstring fix landed correctly — both convenience methods now document the real behavior (`result.exception?.linkException is CancelledException`), matching the tested behavior in `cancellation_test.dart:50-52`.
- `cancellation_test.dart` passes 7/7 clean, and `dart analyze lib test` is clean on the PR head.

Targeted checks I ran on the PR head (`e904569`): `dart analyze lib test` → no issues; `dart test test/cancellation_test.dart` → 7/7 pass; `cancellable_http_link_test.dart` → 13/13 pass in isolation but see the flakiness note below.

---

### Issues & Suggestions

#### 1. Flaky integration test under parallel load (Medium, non-blocking)

`cancellable_http_link_test.dart:368` — "multiple concurrent requests - cancel one, others complete" fails intermittently when run in the full suite (I got `+19 -1` twice) but passes 3/3 in isolation. The failure is at `:419` (`expect(result2.hasException, isFalse)` — the non-cancelled request picks up an exception under load).

Root cause: the test binds a real `HttpServer` with a hardcoded 2s delay for the first request, and when 13 tests spin up real servers concurrently the timing assumptions break. This is a CI hazard — it'll cause intermittent red builds and erode trust in the signal.

Not a blocker for the cancellation feature itself (the isolation tests all pass), but worth fixing before or right after merge. Options: tag this group `@Tags(['integration'])` and run it serially, or replace the real-server concurrency test with a deterministic mock-link version.

---

### Summary

Concept ACK — both gates met, merge-ready. The flaky test is a quick follow-up, not a blocker. Nice work getting this 6-year-old request (#181) across the line!

vincenzopalazzo (approved):

vincenzopalazzo:
Waiting for Ci and merging, thanks!

PR #1522, opened 2026-04-10
fix(graphql): guard query completer when stream emits multiple responses

tomwap:
## Summary

`QueryManager` completed the same `Completer` on every `Response` from the link stream. If the link emitted more than one event (e.g. multiple `Response`s), the second completion threw **Bad state: Future already completed**. This change only completes when the completer is still pending, consistent with the existing `onError` path.

**Why merge:** Fixes a runtime crash for any setup where the link stream can emit multiple values for a single query, and adds a regression test so the first emitted response is used when multiple emissions occur.

#### Reproduction
Observed in the app with a very short `queryRequestTimeout` and `TimedInMemoryStore`, which can surface multiple stream events and trigger **Bad state: Future already completed**:
```dart
return GraphQLClient(
  link: links,
  cache: GraphQLCache(
    store: TimedInMemoryStore(),
  ),
  queryRequestTimeout: const Duration(milliseconds: 1),
);
```

### Breaking changes

None.

#### Fixes / Enhancements

- Fixed `QueryManager` calling `completer.complete` for every stream event; the first response is used when multiple events are emitted.
- Added a unit test that asserts the first emitted `Response` is used when the link stream yields multiple responses.

#### Docs

- N/A.

### DIFF
--- a/packages/graphql/lib/src/core/query_manager.dart
+++ b/packages/graphql/lib/src/core/query_manager.dart
@@ -282,13 +282,19 @@ class QueryManager {
       }
 
       // Listen for the first response or error
-      responseStream.listen(completer.complete,
-          onError: (Object error, StackTrace stackTrace) {
-        if (!completer.isCompleted) {
-          // We return the first error encountered
-          completer.completeError(error, stackTrace);
-        }
-      });
+      responseStream.listen(
+        (Response value) {
+          if (!completer.isCompleted) {
+            completer.complete(value);
+          }
+        },
+        onError: (Object error, StackTrace stackTrace) {
+          if (!completer.isCompleted) {
+            // We return the first error encountered
+            completer.completeError(error, stackTrace);
+          }
+        },
+      );
 
       // Await the response or error
       response = await completer.future;
--- a/packages/graphql/test/query_manager_test.dart
+++ b/packages/graphql/test/query_manager_test.dart
@@ -42,5 +42,31 @@ void main() {
       );
       client.queryManager.refetchQuery<dynamic>(observable.queryId);
     });
+
+    test('uses first response when link emits multiple events', () async {
+      const ping = 'ping';
+
+      when(link.request(any)).thenAnswer(
+        (_) => Stream.fromIterable(<Response>[
+          Response(
+            data: <String, dynamic>{ping: 1},
+            response: <String, dynamic>{},
+          ),
+          Response(
+            data: <String, dynamic>{ping: 2},
+            response: <String, dynamic>{},
+          ),
+        ]),
+      );
+
+      final result = await client.query(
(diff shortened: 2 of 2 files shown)


packages/graphql/lib/src/core/query_manager.dart
```diff
@@ -282,13 +282,19 @@ class QueryManager {
...
 
       // Listen for the first response or error
-      responseStream.listen(completer.complete,
-          onError: (Object error, StackTrace stackTrace) {
-        if (!completer.isCompleted) {
-          // We return the first error encountered
-          completer.completeError(error, stackTrace);
-        }
-      });
+      responseStream.listen(
```
vincenzopalazzo, line 285: Can we stop listening after the first response instead of only guarding `completer.complete`? `query()` is a one-shot operation, so leaving the subscription alive means long-lived or multipart streams keep emitting after the result has already been returned, links never see cancellation, and the `timeout(...)` wrapper can stay active unnecessarily. `responseStream.first` or explicit `StreamSubscription.cancel()` after the first event would fix that. The new test uses `Stream.fromIterable`, so it will not catch this open-stream case.

vincenzopalazzo (changes requested):
Requesting changes for one remaining stream-lifecycle bug. This fixes the double-completion crash, but `query()` still keeps listening after the first response; the inline comment covers the specific issue.

packages/graphql/lib/src/core/query_manager.dart
tomwap: Isn’t adding `responseStream.first`  effectively the same as reverting [PR #1515](https://github.com/zino-hofmann/graphql-flutter/pull/1515)?

vincenzopalazzo: Good point! No, I am not suggesting reverting to `responseStream.first` — that would break the timeout handling from #1515. The Completer pattern is correct.

What I mean is: keep the Completer + guard, but also cancel the subscription after the first event. Something like:

```dart
final subscription = responseStream.listen(
  (Response value) {
    if (!completer.isCompleted) {
      completer.complete(value);
    }
  },
  onError: (Object error, StackTrace stackTrace) {
    if (!completer.isCompleted) {
      completer.completeError(error, stackTrace);
    }
  },
);

response = await completer.future;
await subscription.cancel();
```

This keeps #1515's timeout handling intact while also cleaning up the subscription after `query()` gets its one-shot result. Without the cancel, the listener stays alive for the lifetime of the stream — links never see cancellation, and the timeout wrapper can keep ticking unnecessarily.

The `isCompleted` guard is still needed as a safety net (race between cancel propagation and a pending event), but the cancel is what actually stops the stream producer.

